PR #185, opened 2023-02-08
183 reactliveblock for documentation

Powerplex:
How it's used in stories:
Capture d’écran 2023-02-08 à 17 35 55

How it shows:
https://user-images.githubusercontent.com/2033710/217592999-90c5c753-3e57-45b4-b4cd-e8da8759781f.mov


documentation/helpers/ReactLiveBlock/utils.ts
```diff
@@ -0,0 +1,27 @@
+import copy from 'copy-to-clipboard'
+import { useCallback, useEffect, useState } from 'react'
+
+export function useClipboard(value: string) {
```
andresz1, line 4: Maybe this hook should be a package in the future ? 🤔

andresz1 (approved):

documentation/helpers/ReactLiveBlock/utils.ts
Powerplex: Yep :) @andresin87 suggested it also :) I can be moved to hooks packages when we make them.

Issue #158, opened 2023-02-07
Set up pre-commit strategy


acd02:
Make sure that linters (prettier / eslint) are being run on staged files

PR #206, opened 2023-02-09
feat: add pre-commit linting

acd02:
**TASK**:
address #158

Add lint-staged in order to handle pre-commit linting

### Types of changes
- [x] 🛠️ Tool


arnau-rius (commented):
Should be nice to lint also *.js files if there are any. I think that with this regex will not catch it: `**/*.(m|c)?ts?(x)`

acd02:
True that, JS is dead to me 😅, I'll make another PR to address it 👍

acd02:
@arnau-rius , addressed by #223 ✅

PR #217, opened 2023-02-10
ci: deployments on the workflow

andresin87:
## ci: deployments on the workflow

**TASK**:


arnau-rius (commented):
Why are we doing this manually and not installing the official Vercel integration that has with Github?

Issue #205, opened 2023-02-09
Scaffolding package


Powerplex:

kikoruiz:
* Structure:

```
/packages/utils/scaffolding/
  package.json (@spark-ui/scaffolding)
  bin/
    spark-scaffolding.{ts,js}
```

* Files to be generated:

```
  - src/Component.test.tsx
  - src/Component.tsx
  - src/index.ts
  - stories/Component.stories.mdx
  - stories/Component.stories.tsx
  - .npmigonore
  - package.json
  - tsconfig.json (?)
  - vite.config.ts (?)
```

PR #221, opened 2023-02-13
feat: add cli util to generate new components

kikoruiz:
## FEAT: add cli util to generate new components

**TASK**: #205 

Have available a CLI tool to generate new components. The command suggested is the following one: `npx spark g button -D "My fancy button"`

### Types of changes

- [x] 🛠️ Tool


packages/utils/cli/templates/stories/[Component.stories.tsx].js
```diff
@@ -0,0 +1,16 @@
+import { pascalCase } from 'pascal-case'
```
Comment (acd02) [inline packages/utils/cli/templates/stories/[Component.stories.tsx].js:1]: We don't need the `stories` folder anymore 🧹 🥳

acd02 (approved):

packages/utils/cli/templates/stories/[Component.stories.tsx].js
Comment (Powerplex) [reply packages/utils/cli/templates/stories/[Component.stories.tsx].js:1]: Indeed, package structure is now documented here and everything is flat inside the `src` folder: https://sparkui.vercel.app/?path=/docs/definitions-package-structure--docs

packages/utils/cli/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@spark-ui/cli-utils",
```
Powerplex, line 2: I am wondering if we should not dedicate this package to scaffolding other packages and rename it to something less generic ?

kikoruiz: I think it's nice to have something global like `spark [command]`, we should add each command in the package.json config. What do you think?

packages/utils/cli/templates/stories/[Component.stories.tsx].js
Comment (kikoruiz) [reply packages/utils/cli/templates/stories/[Component.stories.tsx].js:1]: Nice!! I see that it's working fine right now 🥳

packages/utils/cli/package.json
andresin87: What if there is another directory for tool packages called `tools` like `hooks` and `components`. Maybe not initially but if we consider to become an standard in the company this tool will be useful to create brand components following the same pattern

Powerplex: Regarding directory (workspace) we can always update later it won't impact the package name that is published :) 

(What I mean is we can always change it later if it is merged as is)

packages/utils/cli/bin/spark-generate.mjs
```diff
@@ -0,0 +1,56 @@
...
+          .replace('Component', pascalCase(component))
+          .replaceAll(/\[|\]|\.js$/g, ''),
+        content: module.default({ component, description }),
+      }))
+    )
+
+    const filesToWrite = await Promise.all(templateContents)
+
+    Promise.all(filesToWrite.map(writeFile)).then(() => {
+      log('All component files has been properly written!')
```
kikoruiz, line 53:
```suggestion
      log('All component files have been properly written!')
```

PR #234, opened 2023-02-14
feat(use-mounted-state): add useMountedState react hook

andresin87:
**TASK**: 

### Types of changes


packages/hooks/use-mounted-state/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@spark-ui/use-mounted-state",
+  "version": "0.0.1",
+  "description": "",
```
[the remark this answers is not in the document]
andresin87:
```suggestion
  "description": "check the component's mount state",
```

packages/hooks/use-mounted-state/src/useMountedState.tsx
```diff
@@ -0,0 +1,17 @@
+import { useCallback, useEffect, useRef } from 'react'
+
+export function useMountedState(): () => boolean {
+  const mountedRef = useRef<boolean>(false)
```
acd02, line 4:
```suggestion
  const mountedRef = useRef(false)
```

packages/hooks/use-mounted-state/src/useMountedState.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  const get = useCallback(() => mountedRef.current, [])
```
acd02, line 6: Question, why return a function rather than a boolean directly?

andresin87: yes, the intention is to use it in any place calling directly where you need, not infering the state calculated in the resolved hook placement

andresin87: this is taken from react-use library

acd02: 👍 , but do we need to re-rewrite it and expose it then? Couldn't we use the one from the react-use lib instead?

andresin87: this are kind of primitives, we should provide and maintain set of common tools  for repetitive behaviors in components.

andresin87: this is the simplest one, its just to add the useCombinedState, which uses it.

PR #394, opened 2023-03-14
feat(switch): created switch component

Powerplex:
**TASK**: #381 

### Types of changes
- [x] 🧾 Documentation


packages/components/switch/src/Switch.variants.tsx
```diff
@@ -0,0 +1,65 @@
+import { cva, VariantProps } from 'class-variance-authority'
+
+const defaultVariants = {
+  intent: 'primary',
+  size: 'small',
+} as const
+
+export const switchVariants = cva(
```
acd02, line 8: suggestion, maybe rename that to `variants`, or simply `styles`?

package.json
```diff
@@ -42,6 +42,7 @@
     "@commitlint/config-lerna-scopes": "17.4.2",
     "@commitlint/cz-commitlint": "17.4.4",
     "@commitlint/prompt-cli": "17.4.4",
+    "@radix-ui/react-switch": "^1.0.2",
```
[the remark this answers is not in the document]
Powerplex: Actually this one should be `*` because it is in the root package for the local storybook :D I update it.

And I remove the carets on the other.

packages/components/switch/src/Switch.variants.tsx
```diff
@@ -0,0 +1,65 @@
...
+    'cursor-pointer',
+    'rounded-full border-md border-transparent',
+    'transition-colors duration-200 ease-in-out',
+    'radix-disabled:opacity-50 radix-disabled:cursor-not-allowed',
+    'focus:outline-none focus-visible:ring focus-visible:ring-outline-high',
+  ],
+  {
+    variants: {
+      size: {
+        small: ['h-[1rem]', 'w-[2.25rem]'],
```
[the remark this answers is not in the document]
Powerplex: I just discussed this with @andresz1 , we suggest to keep default tailwind spacing in addition to our custom ones. WDYT ? https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale

packages/components/switch/src/Switch.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  'error',
+  'info',
+  'neutral',
+]
+
+export const Colors = () => (
+  <ReactLiveBlock scope={{ Switch }}>
+    {switchColors.map(color => (
+      <div className="flex items-center gap-sm" key={color}>
+        <label htmlFor="primary">{color}</label>
```
arnau-rius, line 59:
```suggestion
        <label htmlFor={color}>{color}</label>
```

packages/components/switch/src/Switch.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Switch name="primary" intent={color} defaultChecked />
```
arnau-rius, line 60:
```suggestion
        <Switch name={color} intent={color} defaultChecked />
```

package.json
```diff
@@ -99,7 +100,7 @@
     "storybook": "7.0.0-beta.62",
     "tailwind-config-viewer": "1.7.2",
     "tailwindcss": "3.2.7",
-    "tailwindcss-radix": "^2.8.0",
+    "tailwindcss-radix": "2.8.0",
```
acd02, line 103: shouldn't we relocate this dependency within the component?

packages/components/switch/src/Switch.variants.tsx
```diff
@@ -0,0 +1,65 @@
...
```
acd02, line 20: se should avoid using arbitrary values, as it may lead to UI inconsistencies if a different base font size is used at the call site

packages/components/switch/src/Switch.variants.tsx
```diff
@@ -0,0 +1,65 @@
...
+    'pointer-events-none',
+    'rounded-full',
+    'shadow-lg',
+    'ring-0',
+    'transform transition-all duration-200 ease-in-out',
+  ],
+  {
+    variants: {
+      size: {
+        small: ['h-[12px]', 'w-[12px]'],
```
acd02, line 58: same here, we should not use arbitrary values

PR #395, opened 2023-03-14
test(cli-utils): add tests to generate cli

kikoruiz:
## test(cli-utils): add tests to generate cli

**TASK**: #315 

Add first integration tests to CLI command `spark generate`

### Types of changes

- [x] 🛠️ Tool
- [x] 🧪 Test


packages/utils/cli/package.json
```diff
@@ -11,9 +11,14 @@
...
   "type": "module",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:adevinta/spark.git",
+    "directory": "packages/utils/cli"
+  },
   "dependencies": {
     "@clack/prompts": "0.6.2",
-    "@spark-ui/theme-utils": "^2.9.0",
+    "@spark-ui/theme-utils": "2",
```
acd02, line 21: shouldn't we use the "latest" keyword to ensure we always have the most up-to-date version?

acd02 (approved):

Powerplex (approved):

packages/utils/cli/package.json
kikoruiz: But, `latest` tag can contain breaking changes... so I think is better to keep the control for major versions (of our own packages), minors and bugfixes will be always autoupdated.

PR #427, opened 2023-03-20
feat(button): setup button component

andresin87:
### Types of changes

- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles


packages/components/button/src/Button.stories.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+<StoryHeading label="Import" />
+
+```js
+import { Button } from "@spark-ui/button"
+```
+
+<StoryHeading label="Props" />
+
+<ArgsTable of={Button} />
```
andresz1, line 28: Normally props are placed in the bottom of the documentation

packages/components/button/src/Button.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+describe('Button', () => {
+  it('should render', () => {
+    // Given
+
+    // When
+    render(<Button>Hello World!</Button>)
+
+    // Then
+    expect(screen.getByText('Hello World!')).toBeInTheDocument()
```
andresz1, line 15: Use `getByRole('button', {name: ''})` instead 🙏🏻

packages/components/button/src/Button.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  it('should trigger click event', async () => {
+    // Given
+    const user = userEvent.setup()
+    const clickEvent = vi.fn()
+
+    // When
+    render(<Button onClick={clickEvent}>Hello World!</Button>)
+    await user.click(screen.getByText('Hello World!'))
```
andresz1, line 25: Same as above

packages/components/button/src/Button.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const user = userEvent.setup()
+    const clickEvent = vi.fn()
+
+    // When
+    render(
+      <Button onClick={clickEvent} disabled>
+        Hello World!
+      </Button>
+    )
+    await user.click(screen.getByText('Hello World!'))
```
andresz1, line 42: Same as above

andresz1, line 42: I would also check for `disabled` attribute

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { PropsWithChildren } from 'react'
+
+import { buttonStyles, type ButtonStylesProps } from './Button.styles'
+
+/** Review: Prop VS ButtonProps */
+interface Props
```
andresz1, line 6: IMO you should export button `Props`

packages/components/button/src/variants/tinted.ts
```diff
@@ -0,0 +1,88 @@
+const tw = <T>(a: T): T => a
```
andresz1, line 1: This function definition is repeated several times. Can we do something regarding that ? I think that is not needed

andresz1 (changes requested):

packages/components/button/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "@spark-ui/button",
+  "version": "1.0.0",
+  "description": "Button component is used to trigger an action or event, such as submitting a form, opening a Dialog, canceling an action, or performing a delete operation.",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
```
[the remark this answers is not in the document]
andresz1: Also I think that react should be a [peerDependency](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/accordion/package.json#L59)

packages/components/button/src/Button.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+  'success',
+  'alert',
+  'danger',
+  'neutral',
+  'surface',
+]
+const designs: ButtonProps['design'][] = ['filled', 'outlined', 'tinted', 'ghost', 'contrast']
+const shapes: ButtonProps['shape'][] = ['rounded', 'square', 'pill']
+
+export const Sizes: StoryFn = () => (
```
andresz1, line 23: I think that you need to handle `args` here or the code documentation won't work. See https://github.com/adevinta/spark/blob/main/packages/components/radio/src/Radio.stories.tsx#L6

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+      intent = 'primary',
+      shape = 'rounded',
+      size = 'md',
+      ...rest
+    },
+    ref
+  ) => {
+    return (
+      <button
+        role="button"
```
andresz1, line 24: Is role needed ?

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+      disabled = false,
+      intent = 'primary',
+      shape = 'rounded',
+      size = 'md',
+      ...rest
+    },
+    ref
+  ) => {
+    return (
+      <button
```
andresz1, line 23: I think that [slot](https://www.radix-ui.com/docs/primitives/utilities/slot) component should be used here. Normally you want anchors look like buttons. Here you have an [example](https://www.radix-ui.com/docs/primitives/utilities/slot#usage)

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const Button = React.forwardRef<HTMLButtonElement, PropsWithChildren<Props>>(
+  (
+    {
+      design = 'filled',
+      disabled = false,
+      intent = 'primary',
+      shape = 'rounded',
+      size = 'md',
+      ...rest
```
andresz1, line 18: Should we use `rest` or `others` ?

packages/components/button/src/Button.test.tsx
andresin87: OMG. Maybe we should change the generator, cuz it's the 2nd time we got the same error

packages/components/button/src/Button.tsx
andresin87: nope

packages/components/button/src/Button.tsx
andresin87: yes. Its a primitive we might code soon

packages/components/button/src/Button.tsx
andresin87: i prefer `otherProps`

andresz1 (approved):

Issue #434, opened 2023-03-21
simplify spark setup - consolidate tw plugin calls and auto-apply options


acd02:
In oder to streamline the Spark setup:
- wrap the call to `tailwindcss-radix` within our own function, ensuring the correct options are applied, rather than requiring consumers to configure it at the call site
- apply the same approach for our own plugins

As a result, Spark users will only need to call a single plugin, rather than multiple ones

PR #435, opened 2023-03-21
feat: simplify spark setup - consolidate tw plugin calls and auto-apply options

acd02:
## feat(tailwind-plugins): wrap tailwindcss-radix call, to auto-apply correct options

**TASK**: #434

Wrap the call to `tailwindcss-radix` within our own function, ensuring the correct options are applied, rather than requiring consumers to configure it at the call site

### Types of changes
- [x] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles


packages/utils/tailwind-plugins/package.json
```diff
@@ -7,7 +7,8 @@
   },
   "main": "index.js",
   "dependencies": {
-    "@spark-ui/theme-utils": "^2.10.0"
+    "@spark-ui/theme-utils": "^2.9.1",
+    "tailwindcss-radix": "^2.8.0"
```
andresin87, line 11: "@spark-ui/theme-utils": "2",
 "tailwindcss-radix": "2.8.0"

fix the versions

PR #478, opened 2023-03-28
build(icons): disabled prebuild temporarly

andresin87:
## TYPE(SCOPE): TITLE

**TASK**:


packages/components/icons/scripts/build.mjs
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
```
andresin87, line 1: 🚨🚨🚨🚨🚨

andresz1 (approved):

PR #557, opened 2023-04-05
chore: cleaning warning messages from StoryBook

soykje:
As we're investigating on Storybook v7 issues, there was some warnings we wanted to clean : missing keys, Canvas usage update, ...

### Types of changes
- [x] 🛠️ Tool
- [x] 🧾 Documentation


soykje:
Still having an issue with
image

andresin87 (approved):

soykje:
> i think it would be resolved with my current version. Probably its cloning an string and it warns there is no ref and other props cannot be provided to a single string

I thought (with @Powerplex) it could be related to icons not having the `forwardRef`... wdyt? Anyway, it doesn't seem necessary I guess to use this component here, as we only need to pass a classname to our icon... which is already ok. :+1:

Issue #554, opened 2023-04-05
Storybook - cache issue blocking loading of components


Powerplex:
Happening often in dev mode, not in production:

![image](https://user-images.githubusercontent.com/66770550/230029134-f6ff2651-fe54-4d66-a399-9c906e900bce.png)

soykje:
I don't understand our Vite setup, neither the purpose of the `react-vite` package... Official documentation suggests to install the builder specifically: `https://storybook.js.org/docs/react/builders/vite` Would be happy to discuss with any expert from the team! :smile:

Powerplex:
It comes from the `@spark-ui/icons` components. For some reasons storybook tries to load hundrers of icons (every icon) from cache, and it fails.

PR #580, opened 2023-04-12
554 storybook cache issue blocking loading of components

kikoruiz:
## FIX: storybook cache issue blocking loading of components

**TASK**: #554

Get rid of the blocking issue in Storybook on the first load 

### Types of changes

- [x] 📷 Demo
- [x] 🧪 Test
- [x] 🧠 Refactor


packages/components/icons/src/index.test.tsx
andresin87, line 1: Why do we want to loose the tests for checking each icons follows the agreement?

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -1,5 +1,5 @@
 import * as CheckboxPrimitive from '@radix-ui/react-checkbox'
-import { Check } from '@spark-ui/icons'
+import { Check } from '@spark-ui/icons/src/icons/Check'
```
Powerplex, line 2: We must not use relative imports from packages, it would only work locally. We must find a way to import from what is generated in the `dist` folder of `icons` (`src/icons` does not exists)

Powerplex (changes requested):

andresin87 (approved):

Issue #584, opened 2023-04-12
Checkbox - label


Powerplex:
Following @andresz1 approach to label on `radio` package, implement label for the checkbox

PR #585, opened 2023-04-12
feat(checkbox): wrap every checkbox with its label

Powerplex:
**TASK**: #584

As agreed upon, input component must come with enforced label using `children` and accessible name.
Applied same logic for `@spark-ui/checkbox` than what has been done by @andresz1 on `@spark-ui/radio-group`

### Types of changes
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles


packages/components/checkbox/src/CheckboxLabel.styles.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export const checkboxLabelStyles = cva(['flex', 'items-center', 'gap-md', 'text-body-1'], {
+  variants: {
+    disabled: {
+      true: ['text-neutral/dim-2', 'cursor-not-allowed'],
+      false: ['cursor-pointer'],
+    },
+  },
+  defaultVariants: defaultCheckboxVariants,
```
acd02, line 14:
```suggestion
  defaultVariants: {
    disabled: false,
  },
```

packages/components/checkbox/src/CheckboxLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+   * The id of the element the label is associated with.
+   */
+  htmlFor?: string
+  /**
+   * When true, prevents the user from interacting with the checkbox item.
+   */
+  disabled?: boolean
+}
+
+export const CheckboxLabel = ({ className, disabled, ...others }: CheckboxLabelProps) => {
```
acd02, line 23:
```suggestion
export const Label = ({ className, disabled, ...others }: CheckboxLabelProps) => {
```

packages/components/checkbox/src/CheckboxLabel.tsx
```diff
@@ -0,0 +1,25 @@
+import { Label } from '@radix-ui/react-label'
+import { PropsWithChildren } from 'react'
+
+import { checkboxLabelStyles, type CheckboxLabelStylesProps } from './CheckboxLabel.styles'
+
+export interface CheckboxLabelProps
```
acd02, line 6:
```suggestion
export interface LabelProps
```

packages/components/checkbox/src/CheckboxInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+   */
+  name?: string
+}
+
+export interface CheckboxInputProps
+  extends CheckboxRadixProps, // Radix props
+    StylesProps, // CVA props (variants)
+    Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'value'> {} // Native HTML props
+
+export const CheckboxInput = React.forwardRef<
```
acd02, line 43:
```suggestion
export const Input = React.forwardRef<
```

packages/components/checkbox/src/CheckboxLabel.styles.ts
Powerplex: I always split it because when we start adding variants with strings instead of boolean, doing it directly in the `cva()` argument generates eslint errors (invalid TW classNames)

PR #593, opened 2023-04-14
refactor: update eslint curly rule

acd02:
Update eslint curly rule

### Types of changes
- [x] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -64,7 +64,7 @@ module.exports = {
     // Misc
     'array-callback-return': 1,
     complexity: [1, 8],
-    curly: [1, 'all'],
+    curly: [1, 'multi-line'],
     'max-lines': [1, { max: 300, skipBlankLines: true, skipComments: true }],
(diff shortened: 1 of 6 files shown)


documentation/helpers/StoryStatus/index.tsx
```diff
@@ -18,9 +18,7 @@ interface Props {
 }
 
 export const StoryStatus = ({ status }: Props): ReactElement | null => {
-  if (status === 'final') {
-    return null
-  }
+  if (status === 'final') return null
```
Powerplex, line 21: I personally don't like this at all.

It will make it hard to find return statement in the middle of files. I think we should be free to do what we want and have no rule for that.

acd02: you are free to do as you like ;)
https://eslint.org/docs/latest/rules/curly#multi-line

kikoruiz (approved):

soykje (approved):

Issue #544, opened 2023-04-03
[Components] Tabs


soykje:
![image](https://user-images.githubusercontent.com/66770550/233297720-b43323d6-8ce4-47b0-8f8f-b6695684acc7.png)

soykje:
Both technical and design reviews on their ways 🚀

turolopezsanabria:
✅ OK design review

turolopezsanabria:
✅ OK design review

turolopezsanabria:
- In story "intent" add labels to each example
- We need lables for the sizes as well.
- The label for States could change to Disabled state
- Update the prop icon to reflect that now are childrens (like chips)
- Add which one is the default in all black.

soykje:
Everything has been updated. Waiting for final reviews!

PR #608, opened 2023-04-17
544 components tabs

soykje:
**TASK**: #544 

### Types of changes
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧪 Test

### Screenshots - Animations
image


packages/components/tabs/src/Tabs.styles.ts
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable tailwindcss/no-custom-classname */
+import { cva, type VariantProps } from 'class-variance-authority'
+
+export const tabsRootStyles = cva([
```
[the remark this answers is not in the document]
acd02: Also, maybe you could have dedicated files instead having all of the styles declared inside this file, i.e:
- TabsContent.styles.ts
- TabsTrigger.styles.ts
- ...

.storybook/main.cjs
```diff
@@ -40,6 +40,24 @@ module.exports = {
   typescript: {
     check: true,
     reactDocgen: 'react-docgen-typescript',
+    reactDocgenTypescriptOptions: {
```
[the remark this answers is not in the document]
soykje: You're right: https://github.com/adevinta/spark/pull/612

packages/components/tabs/src/Tabs.styles.ts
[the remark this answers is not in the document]
soykje: As we have this flat architecture, I found painless to multiply files, also styles ones. Seemed still ok to me here, as there are not so many styles? But if it's really annoying I can split that :+1:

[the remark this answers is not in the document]
acd02: I think we should try to stay consistent,
https://sparkui.vercel.app/?path=/docs/developer-definitions-package-structure--docs

Btw, we also should add the `.styles.ts` to this doc 😅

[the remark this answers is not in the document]
soykje: Allright, just splitted styles :+1:

packages/components/tabs/src/TabsTrigger.styles.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable tailwindcss/no-custom-classname */
```
acd02, line 1: please remove this comment 😅

packages/components/tabs/src/TabsList.styles.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable tailwindcss/no-custom-classname */
```
acd02, line 1: please remove this comment 😅

packages/components/tabs/src/TabsTrigger.tsx
```diff
@@ -0,0 +1,58 @@
...
+    ref
+  ) => {
+    const { intent, size, selectedTab } = useTabsContext()
+    const innerRef = useRef(null)
+
+    const triggerRef = ref || innerRef
+
+    useEffect(() => {
+      if (typeof triggerRef !== 'function' && selectedTab === value) {
+        triggerRef.current?.scrollIntoView({
```
acd02, line 37: Do we really want / need this `scrollIntoView `?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,112 @@
...
+      ...rest
+    },
+    ref
+  ) => {
+    const innerRef = useRef(null)
+    const listRef = ref || innerRef
+    const { orientation, selectedTab, setSelectedTab } = useTabsContext()
+    const tabs = useRef<ReactElement[]>(children)
+
+    const tabPivots = useMemo(() => {
```
acd02, line 39: Since the computation done inside this useMemo does not seem expensive, and `selectedTab` is a dependency, is it really necessary to memoize this function?

packages/components/tabs/src/TabsTrigger.tsx
soykje: What would you suggest?

turolopezsanabria:
✅ OK Design Review

turolopezsanabria (approved):

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+  ) => {
+    const wrapperRef = useRef(null)
+    const innerRef = useRef(null)
+    const listRef = ref || innerRef
+    const { orientation } = useTabsContext()
+
+    const { width } = useResizeObserver(wrapperRef)
+
+    const [withArrows, setWithArrows] = useState<
+      Record<'prev' | 'next', 'visible' | 'hidden' | 'disabled'>
```
andresz1, line 50: Should this be a enum `'visible' | 'hidden' | 'disabled'` ? For example:
```js
useState<{prev: TabListButtonState, next: TabListButtonState}>({
  prev: TabListButtonState.hidden,
  next: TabListButtonState.hidden
})
```

packages/components/tabs/src/TabsTrigger.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const { intent, size, selectedTab } = useTabsContext()
+    const innerRef = useRef(null)
+
+    const triggerRef = ref || innerRef
+
+    useEffect(() => {
+      if (triggerRef && 'current' in triggerRef && triggerRef.current && selectedTab === value) {
+        triggerRef.current.focus()
+      }
+    }, [selectedTab, value, triggerRef])
```
andresz1, line 48: I think that there is not need to add references in dependency array since they don't trigger a rerender 🤔

packages/components/tabs/src/TabsContext.tsx
```diff
@@ -0,0 +1,22 @@
+import type { TabsProps } from '@radix-ui/react-tabs'
```
andresz1, line 1: The context for the radio was organized in a different way https://github.com/adevinta/spark/blob/main/packages/components/radio/src/RadioGroupProvider.tsx. Should we come with agreement regarding this ?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+          <Button
+            shape="square"
+            intent="surface"
+            size="sm"
+            className={navigationArrowStyles()}
+            onClick={handlePrevClick}
+            disabled={withArrows.prev === 'disabled'}
+            aria-label="Scroll left"
+          >
+            <Icon size="sm">
```
andresz1, line 147: `size` should not be needed here I think 🤔

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <div className={wrapperStyles({ className })} ref={wrapperRef}>
+        {withArrows.prev !== 'hidden' && (
+          <Button
+            shape="square"
+            intent="surface"
+            size="sm"
+            className={navigationArrowStyles()}
+            onClick={handlePrevClick}
+            disabled={withArrows.prev === 'disabled'}
+            aria-label="Scroll left"
```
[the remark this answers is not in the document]
andresz1: cc @Powerplex @soykje WDYT ?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    const [withArrows, setWithArrows] = useState<
+      Record<'prev' | 'next', 'visible' | 'hidden' | 'disabled'>
+    >({ prev: 'hidden', next: 'hidden' })
+
+    useEffect(() => {
+      /**
+       * Show/hide arrows
+       */
+      /* istanbul ignore if -- @preserve */
```
acd02, line 57: is this comment needed?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+      if (!('current' in listRef) || !listRef.current) {
+        return
+      }
+
+      if (orientation !== 'horizontal') {
+        setWithArrows({ prev: 'hidden', next: 'hidden' })
+      } else {
```
acd02, line 64: is this `else` needed?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+        })
+      }
+    }, [orientation, listRef, width])
+
+    useEffect(() => {
+      /**
+       * Enable/disable arrows
+       */
+      /* istanbul ignore if -- @preserve */
+      if (!('current' in listRef) || !listRef.current || withArrows.prev === 'hidden' || loop) {
```
acd02, line 77:
```suggestion
      if (typeof listRef === 'function' || !listRef.current || withArrows.prev === 'hidden' || loop) {
```

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+        return
+      }
+
+      const toggleArrowsVisibility = (target: HTMLDivElement) => {
+        setWithArrows({
+          prev: target.scrollLeft > 0 ? 'visible' : 'disabled',
+          next:
+            target.scrollLeft + target.clientWidth < target.scrollWidth ? 'visible' : 'disabled',
+        })
+      }
```
acd02, line 87:
```suggestion
      const toggleArrowsVisibility = (event: Event) => {
       if (!(event.target instanceof HTMLDivElement)) return
       const { target } = event
        setWithArrows({
          prev: target.scrollLeft > 0 ? 'visible' : 'disabled',
          next:
            target.scrollLeft + target.clientWidth < target.scrollWidth ? 'visible' : 'disabled',
        })
      }
```

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+      const currentList = listRef.current
+
+      toggleArrowsVisibility(currentList)
+
+      currentList.addEventListener('scroll', ({ target }) =>
+        toggleArrowsVisibility(target as HTMLDivElement)
+      )
```
acd02, line 95:
```suggestion
currentList.addEventListener('scroll', toggleArrowsVisibility)
```

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+      return () =>
+        currentList.removeEventListener('scroll', ({ target }) =>
+          toggleArrowsVisibility(target as HTMLDivElement)
+        )
```
acd02, line 100:
```suggestion
currentList.addEventListener('scroll', toggleArrowsVisibility)
```

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+    }, [listRef, withArrows.prev, loop])
+
+    const handlePrevClick = useCallback(() => {
+      /* istanbul ignore if -- @preserve */
```
acd02, line 104: What do all those "istanbul ignore" comments mean?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+        toggleArrowsVisibility(target as HTMLDivElement)
+      )
+
+      return () =>
+        currentList.removeEventListener('scroll', ({ target }) =>
+          toggleArrowsVisibility(target as HTMLDivElement)
+        )
+    }, [listRef, withArrows.prev, loop])
+
+    const handlePrevClick = useCallback(() => {
```
acd02, line 103: Is it necessary to wrap this function with a `useCallback`? Does it involve heavy computation?

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+      listRef.current.scrollTo({
+        left: shouldLoopForward
+          ? listRef.current.scrollLeft + listRef.current.scrollWidth - listRef.current.clientWidth
+          : listRef.current.scrollLeft - listRef.current.clientWidth,
+        behavior: 'smooth',
+      })
+    }, [listRef, loop])
+
+    const handleNextClick = useCallback(() => {
```
acd02, line 119: Same question here

packages/components/tabs/src/TabsRoot.tsx
```diff
@@ -0,0 +1,65 @@
...
+    ref
+  ) => {
+    const [selectedTab, setSelectedTab] = useState<TabsContextInterface['selectedTab']>(
+      value || defaultValue
+    )
+
+    const handleChange = (v: string) => {
+      onValueChange?.(v)
+      setSelectedTab(v)
+    }
```
andresin87, line 44:
```suggestion
    const handleChange = (value: string) => {
      onValueChange?.(value)
      setSelectedTab(value)
    }
```

Pressing keys has no cost. Name variables with a deterministic relevant name.

andresin87: `e` can be error or event or ...
`d` can be data or digit or decimal or ...
`v` can ve value or version or ...

packages/components/tabs/src/useResizeObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+type ResizeCallback = (entry?: ResizeObserverEntry) => void
+
+export const useResizeObserver = <T extends HTMLElement>(
+  target: RefObject<T> | T | null,
+  onResize?: ResizeCallback
+): Size => {
+  const [size, setSize] = useState<Size>({ width: undefined, height: undefined })
+  const resizeObserverRef = useRef<ResizeObserver>()
+  const resizeCallbackRef = useRef<ResizeCallback | undefined>(onResize)
+
```
acd02, line 17: see -> https://tkdodo.eu/blog/refs-events-and-escape-hatches
```suggestion
// 👇 update the ref when the callback changes
  React.useEffect(() => {
    resizeCallbackRef.current = onResize
  }, [onResize])
```

packages/components/tabs/src/useResizeObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+      })
+    }
+
+    resizeObserverRef.current.observe(targetElm as unknown as HTMLElement)
+
+    return () => {
+      resizeObserverRef.current &&
+        resizeObserverRef.current.unobserve(targetElm as unknown as HTMLElement)
+    }
+  }, [target, resizeCallbackRef])
```
acd02, line 39:
```suggestion
  }, [target])
```

packages/components/tabs/src/useResizeObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+  const resizeObserverRef = useRef<ResizeObserver>()
+  const resizeCallbackRef = useRef<ResizeCallback | undefined>(onResize)
+
+  useEffect(() => {
+    const targetElm = target && 'current' in target ? target.current : target
+    if (!targetElm) {
+      return
+    }
+
+    if (!resizeObserverRef.current) {
```
acd02, line 24: to avoid a nesting block, maybe switch the condition:
`if (resizeObserverRef.current) return`

src/tailwind.css
```diff
@@ -1,3 +1,12 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer utilities {
```
acd02, line 5: this should be handled by our Tailwind utilities plugin -> `packages/utils/tailwind-plugins/utilities/index.js`

packages/components/tabs/src/TabsList.tsx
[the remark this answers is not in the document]
andresz1: Label translation should be done by spark users. IMO the best way to allow this is providing a compound/slot components API

packages/utils/tailwind-plugins/utilities/index.js
```diff
@@ -12,5 +12,7 @@ module.exports = plugin.withOptions(options => ({ addUtilities, theme }) => {
         'transitionTimingFunction.in'
       )}, border-color ${theme('transitionDuration.300')} ${theme('transitionTimingFunction.in')}`,
     },
+    'no-scrollbar': { scrollbarWidth: 'none' },
+    'no-scrollbar::-webkit-scrollbar': { display: 'none' },
```
acd02, line 16:
```suggestion
    'u-no-scrollbar': {
      scrollbarWidth: 'none',
      'no-scrollbar::-webkit-scrollbar': {
        display: 'none'
      }
    },
```

soykje: This also seems to work, wdyt?
```
'.no-scrollbar': {
      scrollbarWidth: 'none',
      '&::-webkit-scrollbar': { display: 'none' },
    },
```

packages/components/tabs/src/TabsList.styles.ts
```diff
@@ -0,0 +1,20 @@
+import { cva } from 'class-variance-authority'
+
+export const wrapperStyles = cva(['relative flex'])
```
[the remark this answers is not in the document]
soykje: If I'm not wrong, this is to take care of custom styling, as you'll see in implementation:
```tsx
<div className={wrapperStyles({ className })} ref={wrapperRef}>...
```

packages/components/tabs/src/useResizeObserver.ts
acd02: My point was to switch the predicate, not so much the syntax ;)
```tsx
if (resizeObserverRef.current) {
  return
}
```

packages/components/tabs/src/TabsList.styles.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const listStyles = cva([
```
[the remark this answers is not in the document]
soykje: Maybe I missed something... but what would be the point of using `tw`, if we already have `cva` at the same custom (both are already included in the package)?
I find it clearer to always have the same wrapping function to write styles.

packages/components/tabs/src/TabsList.tsx
```diff
@@ -0,0 +1,183 @@
...
+    const [withArrows, setWithArrows] = useState<
+      Record<'prev' | 'next', 'visible' | 'hidden' | 'disabled'>
+    >({ prev: 'hidden', next: 'hidden' })
+
+    useEffect(() => {
+      /**
+       * Show/hide arrows
+       */
+      /* istanbul ignore if -- @preserve */
+      if (!('current' in listRef) || !listRef.current) {
```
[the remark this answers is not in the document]
soykje: Isn't it the same?

packages/utils/tailwind-plugins/utilities/index.js
acd02: It would be better to follow the standard naming convention, `u-*`

packages/components/tabs/src/TabsList.tsx
soykje: This is about coverage.

packages/utils/tailwind-plugins/utilities/index.js
soykje: Wasn't aware of this naming convention, got it! :+1:

packages/components/tabs/src/useResizeObserver.ts
soykje: This is also not allowed... And I'd have to `return undefined`, which seems uglier than current option, don't you think?

packages/components/tabs/src/TabsList.tsx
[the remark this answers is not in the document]
soykje: You're completely right, we should make an agreement on this. The `label` prop seems confusing to me, but we can discuss about this :+1:

packages/components/tabs/src/TabsList.styles.ts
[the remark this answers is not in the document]
acd02: You could do:
```tsx
import { cx } from 'class-variance-authority'

<div className={cx(wrapperStyles(), className} ref={wrapperRef}>...
```

My point was that using `cva` might be a bit overkill in this case, since it's typically used to handle more complex styles. If you're only declaring a few static classes, it might be more appropriate to use a simpler solution 🤷

packages/components/tabs/src/TabsList.tsx
soykje: Yes here it isn't, but seemed clearer to me, just as we declared above, for `Button`, a `size="sm"`. Also, what if default size value for icon change? :thinking:

packages/components/tabs/src/TabsList.tsx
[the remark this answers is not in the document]
soykje: Yes, but why would I do that double typo on purpose?!

packages/components/tabs/src/TabsList.tsx
soykje: For coverage concern

packages/components/tabs/src/useResizeObserver.ts
soykje: If I'm not wrong, this is precisely because of this line, that I didn't need your `useEffect` that updates callback ref.

packages/components/tabs/src/useResizeObserver.ts
soykje: Hmmm... My bad, it seems I was mistaking here :thinking:

soykje: Hmmm sorry I was in a mistake here with the `undefined`. Anyway, the thing is here, I dont' want to exit from function if `resizeObserverRef.current` is defined... I just want here to create a new `ResizeObserver` in case there isn't already one.

packages/components/tabs/src/TabsList.tsx
soykje: Technically not, but clearer to me, to attach my block within the `orientation` condition.

documentation/helpers/ArgTypes/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ArgTypes as StorybookArgTypes } from '@storybook/blocks'
+import type { FC } from 'react'
+
+interface Props<T> {
+  of?: T
+  subcomponents?: Record<string, T>
+}
+
+export const ArgTypes = <T extends FC>({ of, subcomponents = {} }: Props<T>) => {
+  const componentsList = {
```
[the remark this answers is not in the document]
soykje: I kindly disagree here... For me a list is just like a "collection" and kind of _unformatted_... Renaming it would bring unnecessary complexity on a simple naming, to me.

packages/components/tabs/src/TabsList.tsx
soykje: Yes, but I wanted to make the component ready for coverage constraints... Anyway, I can remove these comments if it's a problem :+1:

packages/components/tabs/src/TabsList.tsx
soykje: Could you please explain the win in this change?

packages/components/tabs/src/TabsContext.tsx
soykje: You're right. We may be interested in future refactoring, with a context creator util, that would help us achieve that... wdyt?

andresz1: 👍🏻

packages/components/tabs/src/TabsRoot.tsx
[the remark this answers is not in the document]
soykje: You're completely right! After all those changes on navigation arrows and so I forgot to remove my unnecessary code. Here it is, thx! :+1:

packages/components/tabs/src/useResizeObserver.ts
acd02: > I just want here to create a new ResizeObserver in case there isn't already one.

Yes, so you should exit the function if there is already a set resizeObserver to prevent creating a new one, right?

documentation/helpers/ArgTypes/index.tsx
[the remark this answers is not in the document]
acd02: > Elseway I would have declared specificly something like componentsArray

That is exactly my point, `componentsList` makes you think you are dealing with an array, when it's not the case 😅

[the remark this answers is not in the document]
soykje: I think we're misunderstanding each other :sweat_smile: Maybe we could discuss this specific point directly in Slack, wdyt?

packages/components/tabs/src/TabsList.styles.ts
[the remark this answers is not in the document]
soykje: Maybe I don't get your point, but here in your last proposal, I should use `tw()` and `cx()`, instead of simply `cva()`? Not sure which option is the most overkill here :sweat_smile:

packages/components/tabs/src/useResizeObserver.ts
soykje: If there is already a `resizeObserverRef`, I will use it, not exit the function: https://github.com/adevinta/spark/blob/544-components-tabs/packages/components/tabs/src/useResizeObserver.ts#L33

packages/components/tabs/src/useResizeObserver.ts
soykje: If I'm not wrong, there is no need here, as the callback is a prop. So whenever it changes, the ref also will?

packages/components/tabs/src/useResizeObserver.ts
soykje: But as said, if my prop changes, it will rerender and so keep my ref synced?

packages/components/tabs/src/TabsList.tsx
soykje: @andresz1 I personnally don't like enums... But if you prefer I can use a union for `'visible' | 'hidden' | 'disabled'`?

packages/components/tabs/src/useResizeObserver.ts
acd02: Nope, see -> https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

> with code like this:
> ```tsx
> function Comp() {
>   const logRef = React.useRef(console.log)
>   React.useEffect(() => {
>     logRef.current(new Date())
>   }, [logRef]) // <-- ✅ No eslint warning here
>   return <div>{/* stuff here */}</div>
> }
> ```
> That's because the logRef value can never change. **So it actually makes no difference whether you include it or not** and I guess the authors of the lint rule decided to not bother you about something that doesn't matter. Good call.

To get a better idea of why you need to update your ref, see this [codesandbox](https://codesandbox.io/s/late-wind-ht22gy?file=/src/pages/Playground/index.tsx) I've made that illustrates the problem:
1. increment the counter to to a value greater than 10 by clicking on the button
2. check the logs (you'll notice there is a problem) ❌
3. now, uncomment lines 13-15
4. refresh the view
5. increment the counter to a value greater than 10 by clicking on the button
6. check the logs (things will be working as intended) ✅

packages/components/tabs/src/useResizeObserver.ts
acd02:
```tsx
useEffect(() => {
    if (resizeObserverRef.current) return

    const targetElm = target && 'current' in target ? target.current : target
    if (!targetElm) {
      return
    }

    resizeObserverRef.current = new ResizeObserver(([entry]) => {
      const { inlineSize: width, blockSize: height } = entry?.borderBoxSize[0] ?? {}
      resizeCallbackRef.current?.(entry)

      setSize({ width, height })
    })

    resizeObserverRef.current.observe(targetElm as unknown as HTMLElement)

    return ...
  }, [target, resizeCallbackRef])
```

The code above checks if a resizeObserver is already set, and if so, returns. If not, a new ResizeObserver is created, and its observe method is called on the target element.

vs
```tsx
useEffect(() => {
    const targetElm = target && 'current' in target ? target.current : target
    if (!targetElm) {
      return
    }

    if (!resizeObserverRef.current) {
      resizeObserverRef.current = new ResizeObserver(([entry]) => {
        const { inlineSize: width, blockSize: height } = entry?.borderBoxSize[0] ?? {}
        resizeCallbackRef.current?.(entry)

        setSize({ width, height })
      })
    }

    resizeObserverRef.current.observe(targetElm as unknown as HTMLElement)

    return () => {
      resizeObserverRef.current &&
        resizeObserverRef.current.unobserve(targetElm as unknown as HTMLElement)
    }
  }, [target, resizeCallbackRef])
```

The code above checks if there is no ResizeObserver already set, then create a new one. However, it will always call the `observe` method on it, even if it's not needed.

packages/components/tabs/src/TabsList.styles.ts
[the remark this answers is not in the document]
acd02: Actually, in your case (using only two classes), you could simply do:
```tsx
<div className={cx('relative flex', className} ref={wrapperRef}>...
```

I mean, you don't really need to extract styles when it's only a couple of classes,

Regarding the use of `cva` for static styles, it's more of a subjective matter.
Personally, I tend to use `cva` for dealing with complex styles that have variants. Nevertheless, if you prefer to use `cva` every time for consistency, that's alright with me 👍

packages/components/tabs/src/TabsList.tsx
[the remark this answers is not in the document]
acd02: One rarely makes typos on purpose, and that's my point. By switching to the other syntax, it won't be possible to make a typo, whether it's voluntary or not 😉

Additionally, it saves you the redundancy of having to write `current` twice in the same line. Furthermore, it also spares you from using two `!` operators consecutively, which can result in a code that is slightly less legible, imho

packages/components/tabs/src/TabsList.tsx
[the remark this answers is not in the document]
soykje: As discussed together previously, we'll update the composability (in order to update aria-labels, for eg.) in further PR, after having agreement about how we'll compose our components :+1:

packages/components/tabs/src/useResizeObserver.ts
soykje: Aaaaaaaaaaaaah yes I get it, sorry :sweat_smile:

documentation/helpers/ArgTypes/index.tsx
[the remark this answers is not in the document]
Powerplex: I also would think it is an array and not a record/dictionnary. When I see `componentsList` I think I can do `componentsList.map()` but I can't it would throw an error.

Personally I don't like to specify "array" "record" "list" at the end of my variables, so I would have called it simply `components`. But I also prefer what @acd02 suggested than "componentsList" for an object.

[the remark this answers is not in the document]
soykje: Allright then, changed for `components` instead :+1:

acd02 (approved):

PR #612, opened 2023-04-18
docs: argtypes SB custom component for compound

soykje:
As we're going to work on compound components, we found some issues when using `ArgTypes` from SB. Here is a workaround.

### Types of changes
- [x] 🛠️ Tool
- [x] 🧾 Documentation

### Screenshots - Animations
image

### DIFF
--- a/.storybook/main.cjs
+++ b/.storybook/main.cjs
@@ -40,6 +40,24 @@ module.exports = {
   typescript: {
     check: true,
     reactDocgen: 'react-docgen-typescript',
+    reactDocgenTypescriptOptions: {
+      propFilter: prop => {
+        const prohibitedPropsRegexes = [/node_modules\/\@types\/react\/index.d.ts/]
+
+        if (prop.declarations?.length > 0) {
+          const isProhibitedProps = prop.declarations.some(declaration =>
+            prohibitedPropsRegexes.some(regex => regex.test(declaration.fileName))
+          )
+
+          return !isProhibitedProps
+        }
+
+        return true
+      },
+      componentNameResolver: expression => {
+        return expression.getName()
+      },
+    },
   },
   docs: {
     autodocs: true,
--- a/documentation/helpers/ArgTypes/index.tsx
+++ b/documentation/helpers/ArgTypes/index.tsx
@@ -1,0 +1,29 @@
+import { Tabs } from '@spark-ui/tabs'
+import { ArgTypes as StorybookArgTypes } from '@storybook/blocks'
+
+interface Props {
+  of: any
+  subcomponents?: Record<string, any>
+}
+
+export const ArgTypes = ({ of, subcomponents = [] }: Props) => {
+  const componentsList = {
+    [of.displayName]: of,
+    ...subcomponents,
+  }
+
+  return (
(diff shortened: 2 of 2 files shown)


documentation/helpers/ArgTypes/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const componentsList = {
+    [of.displayName]: of,
+    ...subcomponents,
+  }
+
+  return (
+    <Tabs defaultValue={of.displayName} className="sb-unstyled">
+      <Tabs.List>
+        {Object.entries(componentsList).map(([name]) => (
+          <Tabs.Trigger key={name} value={name} label={name} />
```
acd02, line 19:
```suggestion
        {Object.keys(componentsList).map(componentName) => (
          <Tabs.Trigger key={componentName} value={componentName} label={componentName} />
```

.storybook/main.cjs
```diff
@@ -40,6 +40,24 @@ module.exports = {
...
+
+          return !isProhibitedProps
+        }
+
+        return true
+      },
+      componentNameResolver: expression => {
+        return expression.getName()
+      },
+    },
```
Powerplex, line 60: @andresin87 @andresz1 @kikoruiz @solygambas FYI this is a trick we had to use on the PoC before.

It was the reason why Radix props were not shown in Storybook's ArgsTable :D (or ArgTypes). By default, Storybook exclude all props coming from the `no_modules` folder, where Radix is located :D By doing our custom prop filter (thanks @acd02 ) we only omit react props, not radix props :p 

It might help cleanup how we write props once its merged, maybe its no longer necessary to re-type radix props like we did.

documentation/helpers/ArgTypes/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { Tabs } from '@spark-ui/tabs'
```
Powerplex, line 1: @soykje maybe you need to add dependencies in the root package.json to remove that error ? 

```
"dependencies": {
    "@spark-ui/theme-utils": "*",
    "@spark-ui/tabs": "*", // add this line
  },
```

soykje: Oh yes... Maybe also it wasn't such a good idea to put this new `ArgTypes` in a dedicated PR, as it relies on tabs incoming component? :thinking:

acd02: Maybe we can update this file after the Tab component is merged?

Powerplex: @acd02 that's what I meant when I said I would not do a separate PR for that because we need Tabs first ahah.

What about we introduce a "WIP/Preview" section in storybook ? For components that are not ready to use ? This way we could still merge and iterate ?

We could set `Tabs` version to 0, set its meta to `Alpha/Components/Tabs` (or "Work in progress" instead of alpha) That way in SB it won't be shown with other components.

WDYT ?

acd02: I'm not so fond about that 😅 quick question, can't we use the Tab component from Storybook?

PR #614, opened 2023-04-18
570 icon add auto size to match icon with text size

andresz1:
## Icon auto size

**TASK**: #570 

This PR:
- Add intent prop to `Icon` component. By default `current` value will be set so the current color will be used. No breaking change
- Add `current` value support for `size` prop. This will be the default value using the current font size.
- Add `current` to `spacing` object in the theme since `1em` value was needed
- Change `makeVariants` to accept a partial object. IMO this way is easier to use if not we will need to set all the params we want manually as the second argument

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

Screen Shot 2023-04-18 at 12 41 25


packages/utils/internal-utils/src/variants/index.ts
```diff
@@ -1,12 +1,8 @@
...
 
-/* eslint-disable space-before-function-paren */
-function makeVariants<
-  Variant extends 'design' | 'intent' | 'shape' | 'size',
-  P extends (keyof VariantLookup[Variant])[] = []
->(
-  variants: P extends [] ? VariantLookup[Variant] : Picks<VariantLookup[Variant], P>
-): P extends [] ? VariantLookup[Variant] : Picks<VariantLookup[Variant], P> {
+function makeVariants<Variant extends 'design' | 'intent' | 'shape' | 'size'>(
+  variants: Partial<VariantLookup[Variant]>
```
andresz1, line 4: Let me know what you think @acd02 🙏🏻

packages/components/icon/src/Icon.variants.tsx
```diff
@@ -3,7 +3,18 @@ import { cva, VariantProps } from 'class-variance-authority'
...
+      primary: ['text-primary'],
+      secondary: ['text-secondary'],
+      success: ['text-success'],
+      alert: ['text-alert'],
+      error: ['text-error'],
+      info: ['text-info'],
+      neutral: ['text-neutral'],
+    }),
     size: makeVariants<'size'>({
+      current: ['w-current', 'h-current'],
```
andresz1, line 17: I chose the name `current` instead of `auto` because it will be confused with the auto css value

packages/utils/internal-utils/src/variants/index.ts
acd02: LGTM, it's less cumbersome this way 👍 

But be sure to update every occurrence of that function call 😅

packages/utils/theme/src/defaultTheme.ts
```diff
@@ -199,6 +199,7 @@ export const defaultTheme: Theme = {
   },
   spacing: {
     auto: 'auto',
+    current: '1em',
```
Powerplex, line 202: Adding it to the `spacing` namespace of the tailwind's theme definition might be "too much".

Spacing is used as a basis for the following TW namespaces:

- borderSpacing
- flexBasis
- gap
- height
- inset
- margin
- maxHeight
- padding
- scrollMargin
- scrollPadding
- space
- textIndent
- translate
- width

Adding it to `spacing` means we make it available to all those namespaces, so Spark users would be able to use the following classes:

```
p-current, m-current, gap-current, etc
```

Is it an issue or not ? that is the question :D (All those values will appear in autocomplete of people using Spark)

acd02: Maybe "base"?

acd02: Or we could add a dedicated class just for our use case, something like `.icon-base-size`, `.icon-base` or whatever 😄
WDYT, @Powerplex  @andresz1 ?

```css
.icon-base-size { font-size: 1rem; }
```

acd02: Oh, ok, so then we could do:
```css
.icon-base-size { width: 1em; height: 1em; }
```

andresz1: LGTM @acd02 I will merge this PR since we agreed to change it in another PR

Powerplex: `em` as a unit refers to the font-size. So if you set it as a token inside `spacing`, we add it to all the namespace I listed. So it should be clear by the name that it refers to a unit relative to the font-size (em) or it can be confusing for the users.

"current" is maybe too generic, if we see `p-current`, it is not clear that padding will be 1em (unit related to the base font size of the parent)

Other naming suggestions: 
`p-current-font` OR `p-1em`

(That is if we decide to keep it inside of `spacing`)

acd02: We are going to add a specific utility class to handle that `.icon-base-size` @Powerplex , so we won't be updating our theme

PR #626, opened 2023-04-19
feat(icon-button): add icon button component

andresz1:
## TYPE(SCOPE): TITLE

**TASK**:


packages/components/icon/src/Icon.tsx
```diff
@@ -39,3 +39,5 @@ export function Icon({
     </>
   )
 }
+
+Icon.displayName = 'Icon'
```
andresin87, line 43: we hadnt made any displayName to a component yet till this one

packages/components/icon-button/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "access": "public"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "vite build"
+  },
+  "dependencies": {
+    "@spark-ui/button": "^1.6.0",
```
andresin87, line 15:
```suggestion
    "@spark-ui/button": "1",
```

packages/components/icon-button/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "class-variance-authority": "0.4.0"
```
andresin87, line 16:
```suggestion
    "@spark-ui/button": "1",
    "class-variance-authority": "0.5.2"
```

packages/components/button/src/Button.tsx
```diff
@@ -46,3 +46,5 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     )
   }
 )
+
+Button.displayName = 'Button'
```
andresin87, line 50: If we decide to do that... it must be applied in all packages n the templates

packages/components/icon/src/Icon.tsx
andresz1: Yes! I discussed this with @Powerplex this yesterday. Even though most of the times is not necessary since is inferred from the function name (see https://legacy.reactjs.org/docs/react-component.html#displayname) we noticed that is needed by storybook canvas. I will created another task to set it in all the components and add it to the generator if you all agree

packages/components/button/src/Button.tsx
andresz1: Yes

packages/components/icon-button/package.json
andresz1: I think that for our component @acd02 said that is better to use caret version

andresz1: BTW I think that would be nice to create a wrapper around cva 🤔 WDYT ?

andresin87: why? if its necessary to upgrade the caret version it means its not exactly retrocompatible

acd02: I thought we all agreed that we should use the caret version, whether we choose something like `^1.4.0` or simply `1`, no?

andresin87:
🙏 if you add a display name to a single one, ... spread all that agreement to all of them

packages/components/icon-button/package.json
kikoruiz, line 15: I think we should agree to set only the major versions for our own packages, WDYT?

packages/components/icon-button/src/IconButton.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+)
+
+export const Disabled: StoryFn = _args => (
+  <IconButton aria-label="Disabled button" disabled>
+    {icon}
+  </IconButton>
+)
+
+export const Intent: StoryFn = _args => (
+  <div className="gap-md flex flex-row">
```
kikoruiz, line 67: I guess `flex-row` is not needed, `row` is the default value for `flex-direction`, right?

kikoruiz (approved):
Nice for me, I just put some small comments

PR #646, opened 2023-04-21
feat(text-input): SPIKE

andresin87:
## feat(text-input): Spike

**TASK**: 

I have followed the same pattern as checkbox and radio input API:
```
<InputElement>
  labelfortahtelement
<InputElement>
```
means:
```
<label>
  <input/>
</label>
```

- TextInput: easy add inputs
- Input: html-input and its span

First attempt to manage inputs from spark

### Types of changes


packages/components/text-input/src/Input.tsx
```diff
@@ -0,0 +1,109 @@
...
+export const Input = forwardRef<HTMLInputElement, InputProps>(
+  (
+    { className, mandatory, htmlSize, intent, placeholder, children, onValueChange, ...props },
+    forwardedRef
+  ) => {
+    const [value, setValue] = useCombinedState(props.value, props.defaultValue)
+    const onChangeHandler = (event: ChangeEvent<HTMLInputElement>) => {
+      const value = event.target?.value?.slice(0, htmlSize)
+      setValue(value)
+      typeof onValueChange === 'function' && onValueChange(value)
```
andresz1, line 55: Maybe would be better to call `onValueChange` with the event 🤔 ?

packages/components/text-input/src/InputField.styles.tsx
```diff
@@ -0,0 +1,40 @@
...
+    'border-md',
+    'border-current',
+    'items-center',
+    // 'overflow-hidden'
+  ],
+  {
+    variants: {
+      /**
+       * Color scheme of the inputField.
+       */
```
andresz1, line 19: Is this comment needed ?

packages/components/text-input/src/InputField.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const InputField = forwardRef<HTMLElement, InputFieldProps>(
+  ({ asChild, intent, disabled, className, children, ...props }, forwardedRef) => {
+    const Element = asChild ? Slot : 'span'
+
+    return (
+      <Element
+        ref={forwardedRef}
+        aria-disabled={disabled}
```
andresz1, line 21: Is this needed ?

packages/components/text-input/src/Input.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const Input = forwardRef<HTMLInputElement, InputProps>(
+  (
+    { className, mandatory, htmlSize, intent, placeholder, children, onValueChange, ...props },
+    forwardedRef
+  ) => {
+    const [value, setValue] = useCombinedState(props.value, props.defaultValue)
+    const onChangeHandler = (event: ChangeEvent<HTMLInputElement>) => {
+      const value = event.target?.value?.slice(0, htmlSize)
```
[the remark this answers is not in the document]
andresin87: the default size of an input is 25ch width unless you configure an htmlSize which means you determine how much characters should input had (html size native attribute)

packages/components/text-input/src/Input.tsx
andresin87: I'm following what we have in checkbox and radio for the onValueChange event handler responses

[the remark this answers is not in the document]
andresin87: it's just for if it exists, you mean...

```suggestion
     onValueChange && onValueChange(value)
```
just for checking if its `!== undefined` cuz the type is handled by Ty?

PR #650, opened 2023-04-21
feat(stack): stack component

Powerplex:
**TASK**: #649 

Daily while writing stories, and more recently during workshops with our users, I saw that we often have the need to stack a list of items together with spacing inbetween. It can be done with Tailwind classes, but I think it would be more user-friendly to have a component to manage that behaviour. WDYT ?

### Types of changes
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧪 Test
- [x] 💄 Styles

### Screenshots - Animations
Capture d’écran 2023-04-21 à 19 22 02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9352,6 +9352,10 @@
       "resolved": "packages/components/spinner",
       "link": true
     },
+    "node_modules/@spark-ui/stack": {
+      "resolved": "packages/components/stack",
+      "link": true
+    },
     "node_modules/@spark-ui/switch": {
       "resolved": "packages/components/switch",
       "link": true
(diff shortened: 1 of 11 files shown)


turolopezsanabria:
May this one be related to a component `group` too?
In the visual description of this ticket, I see 2 things:

- a group of a single component (chips grouped together, a button group, etc)
- a stack of different components (login inputs + a button)

andresz1:
In my opinion, the main issue with this is the lack of a responsive API. Additionally, if I'm not mistaken we will be adding all these classes to the generated css even if they are not truly used

acd02:
I don't think such a component is necessary when using TW, as it's quite trivial to mimic what a Stack component does by adding a couple of classes, e.g `<div class="flex gap-md">...</>`, and you're good to go

acd02:
> add some examples of how a Stack could be created with tailwind in the future

yes, we could add a "tips" section eventually 👍

Powerplex:
Thanks for the feedback :) I agree I'm closing it for now.

Issue #540, opened 2023-04-03
[Components] Button - icons (left/right)


andresz1:
### Description

Add icon support to the button component

### Problem Statement/Justification

At this moment we can not add icons to the buttons easily. We should provide a way to do it. The button can have a right and/or a left icon and the size of them should be synchronized with the size of the button.

### Proposed Solution or API

```jsx
<>
  <Button leftIcon={<EmailIcon />}>
    Email
  </Button>
  <Button rightIcon={<ArrowForwardIcon />}>
    Call us
  </Button>
</>
```

Powerplex:
@andresz1 for now maybe we can continue with composability instead of leftIcon/rightIcon ?

Because if we do it for the Button, why not doing it also for the Radio, Checkbox, etc ? (consistency)

PR #659, opened 2023-04-24
docs(button): add left and right icons example

andresz1:
## Button icons

**TASK**: #540 

This PR:
- Add a example of how icons could be added to a button (left, right and both)

### Types of changes
- [ ] 🛠️ Tool
- [x] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

Screen Shot 2023-04-24 at 16 00 14

### DIFF
--- a/packages/components/button/src/Button.stories.tsx
+++ b/packages/components/button/src/Button.stories.tsx
@@ -1,3 +1,5 @@
+import { Icon } from '@spark-ui/icon'
+import { Check, FavoriteOutline } from '@spark-ui/icons'
 import { Meta, StoryFn } from '@storybook/react'
 import { type ComponentProps } from 'react'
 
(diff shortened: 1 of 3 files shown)


andresin87 (approved):

packages/components/button/src/Button.stories.tsx
```diff
@@ -73,8 +75,34 @@ export const Intent: StoryFn = _args => (
...
+export const Icons: StoryFn = _args => (
+  <div className="gap-md flex flex-wrap">
+    <Button>
+      Button
+      <Icon>
+        <FavoriteOutline />
+      </Icon>
+    </Button>
+    <Button>
+      <Icon>
```
andresz1, line 87: Not sure if we should provide a `Button.Icon` component for this. It will be just a wrapper of the `Icon`

packages/components/button/src/Button.stories.tsx
```diff
@@ -64,7 +66,7 @@ export const Design: StoryFn = _args => (
 )
 
 export const Intent: StoryFn = _args => (
-  <div className="gap-md flex flex-row">
+  <div className="gap-md flex flex-wrap">
```
Powerplex, line 69: I have a warning in my editor that `flex-wrap` is deprecated and we should use `wrap` simply :)

packages/components/button/src/Button.stories.tsx
Powerplex: I don't think it's necessary but I'm not opposed to it either.

PR #663, opened 2023-04-25
fix: global fix for custom styles and cva

soykje:
We forgot sometimes to pass the `className` prop to our cva functions. Leading custom styles to break default ones...

### Types of changes
- [x] 💄 Styles

### DIFF
--- a/packages/components/checkbox/src/Checkbox.tsx
+++ b/packages/components/checkbox/src/Checkbox.tsx
@@ -9,8 +9,8 @@ export type CheckboxProps = InputProps
 export const Checkbox = React.forwardRef<
   React.ElementRef<typeof CheckboxPrimitive.Root>,
   CheckboxProps
->(({ children, ...props }, ref) => (
-  <Label data-spark-component="checkbox" disabled={props.disabled}>
+>(({ children, className, ...props }, ref) => (
+  <Label data-spark-component="checkbox" className={className} disabled={props.disabled}>
     <Input ref={ref} {...props} />
     {children}
   </Label>
--- a/packages/components/radio/src/Radio.tsx
+++ b/packages/components/radio/src/Radio.tsx
@@ -17,7 +17,7 @@ export const Radio = forwardRef<HTMLButtonElement, RadioProps>(
     const disabled = disabledProp || context.disabled
 
     return (
-      <RadioLabel disabled={disabled}>
+      <RadioLabel className={className} disabled={disabled}>
         <RadioInput ref={ref} intent={intent} size={size} {...others} />
         {children}
       </RadioLabel>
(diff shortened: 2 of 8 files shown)


packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -9,8 +9,8 @@ export type CheckboxProps = InputProps
 export const Checkbox = React.forwardRef<
   React.ElementRef<typeof CheckboxPrimitive.Root>,
   CheckboxProps
->(({ children, ...props }, ref) => (
-  <Label data-spark-component="checkbox" disabled={props.disabled}>
+>(({ children, className, ...props }, ref) => (
+  <Label data-spark-component="checkbox" className={className} disabled={props.disabled}>
```
soykje, line 13: @Powerplex Is this ok to pass `className` in label, not input?

packages/components/radio/src/Radio.tsx
```diff
@@ -17,7 +17,7 @@ export const Radio = forwardRef<HTMLButtonElement, RadioProps>(
     const disabled = disabledProp || context.disabled
 
     return (
-      <RadioLabel disabled={disabled}>
+      <RadioLabel className={className} disabled={disabled}>
```
soykje, line 20: @andresz1 Is this ok to pass `className` in label, not input?

packages/components/switch/src/Switch.tsx
```diff
@@ -6,9 +6,9 @@ import { Label } from './SwitchLabel'
 export type SwitchProps = InputProps
 
 export const Switch = React.forwardRef<HTMLButtonElement, SwitchProps>(
-  ({ value = 'on', size = 'md', children, ...rest }, ref) => {
+  ({ value = 'on', size = 'md', children, className, ...rest }, ref) => {
     return (
-      <Label data-spark-component="switch" disabled={rest.disabled}>
+      <Label data-spark-component="switch" disabled={rest.disabled} className={className}>
```
soykje, line 11: Same here, @Powerplex

packages/components/checkbox/src/Checkbox.tsx
Powerplex: That's a good question. I think so yes. But it means we can add styles to the whole block not the checkbox itself (but maybe this is good).

For compound components we won't have this question because every subComponent will be able to receive its own `className` :) Let's keep it like you did.

Powerplex (approved):

packages/components/checkbox/src/Checkbox.tsx
Powerplex: If in the future if someone need to pass different styles to the label wrapping the component VS **only on the input**, we might need to introduce a dedicated prop for it.

andresz1 (approved):

packages/components/radio/src/Radio.tsx
andresz1: Yes! I think that is ok

PR #695, opened 2023-04-28
badge component

kikoruiz:
## feat(badge): create badge component

**TASK**: #654 

Create Badge component regarding the documentation here: https://spark.adevinta.com/1186e1705/p/8711ec-badge/b/98915d

### Types of changes

- [ ] 🛠️ Tool
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧪 Test
- [ ] 🧠 Refactor
- [x] 💄 Styles

### DIFF
--- a/packages/components/badge/package.json
+++ b/packages/components/badge/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "@spark-ui/badge",
+  "version": "1.0.0",
+  "description": "Badge component is a visual indicator for numeric values such as tallies and scores.",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "vite build"
+  },
+  "peerDependencies": {
+    "react": "^16.8 || ^17.0 || ^18.0",
+    "react-dom": "^16.8 || ^17.0 || ^18.0",
+    "tailwindcss": "^3.0.0"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:adevinta/spark.git",
+    "directory": "packages/components/badge"
+  },
+  "devDependencies": {
+    "@spark-ui/radio": "1"
+  }
+}
--- a/packages/components/badge/src/Badge.stories.tsx
+++ b/packages/components/badge/src/Badge.stories.tsx
@@ -1,0 +1,56 @@
+import { Radio, RadioGroup } from '@spark-ui/radio'
+import { Meta, StoryFn } from '@storybook/react'
+import { type ComponentProps } from 'react'
+
+import { Badge } from '.'
+
+type BadgeProps = ComponentProps<typeof Badge>
+
+const meta: Meta<typeof Badge> = {
+  title: 'Components/Badge',
+  component: Badge,
+}
+const sizes: BadgeProps['size'][] = ['sm', 'md']
+const intents: BadgeProps['intent'][] = ['error', 'info']
+
+const fakeAvatar = <div className="w-sz-40 h-sz-40 bg-outline rounded-lg" />
+
+export const Default: StoryFn = _args => <Badge>{fakeAvatar}</Badge>
+export const Standalone: StoryFn = _args => <Badge />
+export const StandaloneWithCount: StoryFn = _args => <Badge count={1} />
+export const StandaloneWithinAnotherElement: StoryFn = _args => (
+  <RadioGroup>
+    <Radio value="foo">
+      Foo <Badge size="sm" count={3} />
+    </Radio>
+  </RadioGroup>
+)
+export const RegularCount: StoryFn = _args => <Badge count={1}>{fakeAvatar}</Badge>
+export const LargeCount: StoryFn = _args => <Badge count={1000}>{fakeAvatar}</Badge>
+export const LargeCustomCount: StoryFn = _args => (
+  <Badge count={1000} overflowCount={999}>
+    {fakeAvatar}
+  </Badge>
+)
+
+export const Intents: StoryFn = _args => (
+  <div className="gap-md flex flex-wrap">
+    {intents.map(intent => (
+      <Badge key={intent} intent={intent} count={1}>
+        {fakeAvatar}
+      </Badge>
+    ))}
+  </div>
+)
+
+export const Sizes: StoryFn = _args => (
+  <div className="gap-md flex flex-wrap">
+    {sizes.map(size => (
+      <Badge key={size} size={size} count={25}>
+        {fakeAvatar}
+      </Badge>
+    ))}
+  </div>
+)
+
+export default meta
--- a/packages/components/badge/src/Badge.test.tsx
+++ b/packages/components/badge/src/Badge.test.tsx
@@ -1,6 +1,5 @@
 import { render, screen } from '@testing-library/react'
-import userEvent from '@testing-library/user-event'
-import { describe, expect, it, vi } from 'vitest'
+import { describe, expect, it } from 'vitest'
 
 import { Badge } from './Badge'
 
--- a/packages/components/badge/src/Badge.test.tsx
+++ b/packages/components/badge/src/Badge.test.tsx
@@ -1,0 +1,13 @@
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { describe, expect, it, vi } from 'vitest'
+
+import { Badge } from './Badge'
+
+describe('Badge', () => {
+  it('should render', () => {
+    render(<Badge>Hello World!</Badge>)
+
+    expect(screen.getByText('Hello World!')).toBeInTheDocument()
+  })
+})
--- a/packages/components/badge/src/Badge.tsx
+++ b/packages/components/badge/src/Badge.tsx
@@ -2,21 +2,19 @@ import { forwardRef, PropsWithChildren } from 'react'
 
 import { BadgeItem, BadgeItemProps } from './BadgeItem'
 
-export type BadgeProps = PropsWithChildren<BadgeItemProps>
+export type BadgeProps = PropsWithChildren<Omit<BadgeItemProps, 'type'>>
 
-export const Badge = forwardRef<HTMLDivElement | HTMLSpanElement, BadgeProps>(
-  ({ children, className, ...props }, ref) => {
-    const isStandalone = !children
+export const Badge = forwardRef<HTMLElement, BadgeProps>(({ children, className, ...props }) => {
+  const isStandalone = !children
 
-    return isStandalone ? (
-      <BadgeItem type="standalone" {...props} />
-    ) : (
-      <div ref={ref} className="relative flex">
-        {children}
-        <BadgeItem {...props} />
-      </div>
-    )
-  }
-)
+  return isStandalone ? (
+    <BadgeItem type="standalone" {...props} />
+  ) : (
+    <div className="relative flex">
+      {children}
+      <BadgeItem {...props} />
+    </div>
+  )
+})
 
 Badge.displayName = 'Badge'
--- a/packages/components/badge/src/Badge.tsx
+++ b/packages/components/badge/src/Badge.tsx
@@ -1,0 +1,22 @@
+import { forwardRef, PropsWithChildren } from 'react'
+
+import { BadgeItem, BadgeItemProps } from './BadgeItem'
+
+export type BadgeProps = PropsWithChildren<BadgeItemProps>
+
+export const Badge = forwardRef<HTMLDivElement | HTMLSpanElement, BadgeProps>(
+  ({ children, className, ...props }, ref) => {
+    const isStandalone = !children
+
+    return isStandalone ? (
+      <BadgeItem type="standalone" {...props} />
+    ) : (
+      <div ref={ref} className="relative flex">
+        {children}
+        <BadgeItem {...props} />
+      </div>
+    )
+  }
+)
+
+Badge.displayName = 'Badge'
--- a/packages/components/badge/src/BadgeItem.styles.tsx
+++ b/packages/components/badge/src/BadgeItem.styles.tsx
@@ -1,6 +1,8 @@
 import { makeVariants } from '@spark-ui/internal-utils'
 import { cva, type VariantProps } from 'class-variance-authority'
 
+import { DEFAULT_INTENT, DEFAULT_SIZE, DEFAULT_TYPE } from './config'
+
 export const styles = cva(
   ['inline-flex h-fit', 'empty:p-none', 'text-center font-bold', 'border-md rounded-full'],
   {
@@ -13,10 +15,10 @@ export const styles = cva(
         sm: ['text-small', 'px-[var(--sz-6)] py-[var(--sz-2)]', 'empty:h-sz-12 empty:w-sz-12'],
         md: ['text-caption', 'px-md py-sm', 'empty:h-sz-16 empty:w-sz-16'],
       }),
-      type: makeVariants<'type', ['relative', 'standalone']>({
+      type: {
         relative: ['-left-lg empty:-left-md relative'],
         standalone: [],
-      }),
+      },
     },
     compoundVariants: [
       {
@@ -31,9 +33,9 @@ export const styles = cva(
       },
     ],
     defaultVariants: {
-      intent: 'error',
-      size: 'md',
-      type: 'relative',
+      intent: DEFAULT_INTENT,
+      size: DEFAULT_SIZE,
+      type: DEFAULT_TYPE,
     },
   }
 )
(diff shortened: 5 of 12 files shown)


packages/components/badge/src/BadgeItem.styles.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const styles = cva(
+  ['inline-flex h-fit', 'empty:p-none', 'text-center font-bold', 'border-md rounded-full'],
+  {
+    variants: {
+      intent: makeVariants<'intent', ['error', 'info']>({
+        error: ['bg-error', 'text-on-error'],
+        info: ['bg-info', 'text-on-info'],
+      }),
+      size: makeVariants<'size', ['sm', 'md']>({
+        sm: ['text-small', 'px-[var(--sz-6)] py-[var(--sz-2)]', 'empty:h-sz-12 empty:w-sz-12'],
```
kikoruiz, line 15: I couldn't make it work with the standard padding sizes 😢

Powerplex: Then I think either:
- the mockups should be updated
- OR we were supposed to use width/height instead of padding

packages/components/badge/src/Badge.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { BadgeItem, BadgeItemProps } from './BadgeItem'
+
+export type BadgeProps = PropsWithChildren<Omit<BadgeItemProps, 'type'>>
+
+export const Badge = forwardRef<HTMLElement, BadgeProps>(({ children, className, ...props }) => {
+  const isStandalone = !children
+
+  return isStandalone ? (
+    <BadgeItem type="standalone" {...props} />
```
Powerplex, line 11: `className` must be forwarded to the JSX :)

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,51 @@
+import { forwardRef, HTMLAttributes, PropsWithRef } from 'react'
+
+import { styles, type StylesProps } from './BadgeItem.styles'
+import { DEFAULT_INTENT, DEFAULT_OVERFLOW_COUNT, DEFAULT_SIZE, DEFAULT_TYPE } from './config'
```
andresz1, line 4: IMO there is no need for a `config` file

packages/components/badge/src/Badge.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ) : (
+    <div className="relative flex">
+      {children}
+      <BadgeItem {...props} />
+    </div>
```
andresz1, line 16: - What If I want to add margin to the wrapper for example ?
- `className` not being used

packages/components/badge/src/BadgeItem.tsx
kikoruiz: I think there is a need since there are some default values shared across different files.

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { styles, type StylesProps } from './BadgeItem.styles'
+import {
+  DEFAULT_INTENT,
+  DEFAULT_LABELS,
+  DEFAULT_OVERFLOW_COUNT,
+  DEFAULT_SIZE,
+  DEFAULT_TYPE,
+} from './config'
+
+interface ACCESSIBILITY_LABELS {
```
andresz1, line 12: IMO interfaces should be `CamelCase` and written in singular

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+   */
+  count?: number
+  /**
+   * Maximum numeric value to be dispayed as a count value.
+   */
+  overflowCount?: number
+  /**
+   * Custom labels for accessibility purposes than can be set for each count case.
+   */
+  labels?: ACCESSIBILITY_LABELS
```
kikoruiz, line 31: Not sure it's the best approach for managing labels for a11y... any suggestions?

andresz1: Maybe a `label` prop. The prop should be a function that takes a `string` (parsed count) a returns a `string` that will be used as `aria-label`. This way users can use they i18n library to do the pluralization magic

andresz1: Perhaps role status is needed (see https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22 and https://stackoverflow.com/questions/68170589/do-html-elements-like-pill-badge-need-extra-aria-role)

kikoruiz: > Maybe a `label` prop. The prop should be a function that takes a `string` (parsed count) a returns a `string` that will be used as `aria-label`. This way users can use their favourite i18n library to do the formatting

Yep, I was thinking on that solution, but that way users will need to build on their own the logic behind the design specs we have in ZH right now, not perfect solution I guess...

packages/components/badge/src/BadgeItem.tsx
kikoruiz: Yep, I treated it as a constant, but I think we all should agree on it and write the conclusions somewhere...

packages/components/badge/src/BadgeItem.styles.tsx
kikoruiz: It should be the first option, because working with width is not valid for this component since it can grow depending on its inner content (1, 99, 99+...). Anyway, I'll check it with design/UX after all the technical code review...

packages/components/badge/src/Badge.tsx
kikoruiz: Second point is fixed. First one... I'm not sure, it could be handled in different ways... but we can add a `wrapperClassName` prop just in case.

packages/components/badge/src/BadgeItem.tsx
kikoruiz: Role added ;)

packages/components/badge/src/Badge.tsx
andresz1: I think that would be better to provide a compound component API
```jsx
<BadgeContainer className="">
  <Avatar />
  <Badge className="" />
</BadgeContainer>
```

packages/components/badge/src/BadgeItem.tsx
andresz1: But this would only apply for the `label` and the function is going to provide the parsed count (for example, `+99` and perhaps) so user can do something like this:
```
<Badge label={count => i18n.t("key", {count})} />
```

packages/components/badge/src/BadgeItem.styles.tsx
```diff
@@ -0,0 +1,43 @@
...
+        md: ['text-caption', 'px-md py-sm', 'empty:h-sz-16 empty:w-sz-16'],
+      }),
+      type: {
+        relative: ['-left-lg empty:-left-md relative'],
```
Powerplex, line 19: I'm just wondering if the relative positionning shouldn't use 50% translate insted of arbitrary values (lg, md, etc). Because like you said, the width of the component can change depending on its contents (1, 99, 99+, etc)

WDYT ?

Powerplex (approved):

packages/components/badge/src/BadgeItem.styles.tsx
```diff
@@ -0,0 +1,43 @@
...
+        size: 'md',
+        type: 'relative',
+        class: ['-top-lg empty:-top-sm'],
+      },
+    ],
+    defaultVariants: {
+      intent: DEFAULT_INTENT,
+      size: DEFAULT_SIZE,
+      type: DEFAULT_TYPE,
+    },
```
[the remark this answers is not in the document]
kikoruiz: I'll leave them for now, and we can recheck it later

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+      className,
+      ...others
+    },
+    ref
+  ) => {
+    const hasOverflow = count && count > overflowCount
+    const regularCountLabel =
+      count && count > 1 ? label.pluralCount?.replace('%COUNT%', `${count}`) : label.singleCount
+    const countLabel = hasOverflow
+      ? label.moreThanOverflowCount?.replace('%OVERFLOW_COUNT%', `${overflowCount}`)
```
Powerplex, line 56: I think labels placeholders (ex: `%OVERFLOW_COUNT%`) should not appear in Spark components.

1. It's not visible by consumers through documentation or the component's API (or Typescript intellisense).
2. It is not necessary because the developers  are the one passing the count. So they can simply do `pluralCount: `You have ${count} notifications`

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+   */
+  count?: number
+  /**
+   * Maximum numeric value to be dispayed as a count value.
+   */
+  overflowCount?: number
+  /**
+   * Object containing custom labels for accessibility purposes than can be set for each count case.
+   */
+  label?: AccessibilityLabel
```
Powerplex, line 31: `label` as it is implemented seems too complicated for a DS.

1. `label` implies a string to me, not an object.
2. `label` implies it is an html label, not an aria-label.
3. Most DS keep things simple by simply opening an `aria-label` prop (string), the logic is then managed by them when they implement the component.

packages/components/badge/src/BadgeItem.tsx
Powerplex: Type: string | ({ [INTERNAL_DATA]: string }) => string

<Badge aria-label={({ count, overflowCount }) => i18n.t("key", {count})} />

packages/components/badge/src/Badge.tsx
kikoruiz: Removed `wrapperClassName` prop, it's not mandatory for this iteration, moreover it's difficult to achieve it using a compound component pattern

packages/components/badge/src/Badge.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    render(<Badge count={100} />)
+
+    expect(screen.getByText('99+')).toBeInTheDocument()
+  })
+
+  it('should print properly the aria-label when count is 100', () => {
+    render(
+      <Badge
+        count={100}
+        aria-label={({ overflowCount }) => `More than ${overflowCount} notifications`}
```
andresz1, line 58: Maybe could be nice to add a test for when `aria-label` is a string 🤔

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+      overflowCount = DEFAULT_OVERFLOW_COUNT,
+      'aria-label': label,
+      className,
+      ...others
+    },
+    ref
+  ) => {
+    const hasOverflow = count && count > overflowCount
+    const ariaLabel = typeof label === 'function' ? label({ count, overflowCount }) : label
+    const props = { ...others, ...(ariaLabel && { 'aria-label': ariaLabel }) }
```
andresz1, line 46:
```js
const props = { ...others, 'aria-label': ariaLabel }
```
Is not the same ?

packages/components/badge/src/BadgeItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+   * Maximum numeric value to be dispayed as a count value.
+   */
+  overflowCount?: number
+  /**
+   * A custom label for accessibility purposes. It can also be defined as a builder function
+   * to handle dynamic inner data to create a custom label.
+   */
+  'aria-label'?:
+    | string
+    | (({ count, overflowCount }: { count?: number; overflowCount?: number }) => void)
```
andresz1, line 23: The return type of the function should be a `string`
```js
(({ count, overflowCount }: { count?: number; overflowCount?: number }) => string)
```

packages/components/badge/src/BadgeItem.styles.tsx
andresz1: I think that we should add sizes for `padding`, `margin`, etc. When I tried spark for building a website it was a handicap

packages/components/badge/src/BadgeItem.tsx
andresz1: I think that we should come with an agreement regarding this because it's the first component to have a config file. I would try to avoid generic files like `config`, `utils` files as much as possible.

packages/components/badge/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@spark-ui/badge",
+  "version": "1.0.0",
```
andresz1, line 3: Is something is missing I think that would be better to mark it as unstable

andresz1 (approved):

andresin87 (approved):

packages/components/badge/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:adevinta/spark.git",
+    "directory": "packages/components/badge"
+  },
+  "devDependencies": {
+    "@spark-ui/icon": "1",
+    "@spark-ui/icons": "1",
+    "@spark-ui/radio": "1"
+  }
```
andresin87, line 28:
```suggestion
  },
  "bugs": {
    "url": "https://github.com/adevinta/spark/issues"
  },
  "homepage": "https://sparkui.vercel.app",
  "license": "MIT"
```

packages/components/badge/src/BadgeItem.tsx
kikoruiz: Yep, it's redundant...

packages/components/badge/src/BadgeItem.tsx
kikoruiz: Well, in fact it's a config exporting only constants... but, yep let's make an agreement on this

packages/components/badge/src/Badge.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+  <RadioGroup>
+    <Radio value="foo">
+      Foo <Badge size="sm" count={3} aria-label={({ count }) => `${count} notifications`} />
+    </Radio>
+  </RadioGroup>
+)
+export const WrappingAnIcon: StoryFn = _args => (
+  <Badge size="sm" count={3}>
+    {/** TODO: Check the real behaviour with icons in order to avoid custom styles on them. **/}
+    <Icon size="lg" className="mx-md">
```
andresin87, line 33: This class is adding unnecessary margins to the badge which is confusing. I think its better as:
```suggestion
    <Icon size="lg">
```

packages/components/badge/src/BadgeItem.styles.tsx
```diff
@@ -0,0 +1,31 @@
+import { makeVariants } from '@spark-ui/internal-utils'
+import { cva, type VariantProps } from 'class-variance-authority'
+
+import { DEFAULT_INTENT, DEFAULT_SIZE, DEFAULT_TYPE } from './config'
+
+export const styles = cva(
+  ['inline-flex h-fit', 'empty:p-none', 'text-center font-bold', 'rounded-full outline outline-2'],
+  {
+    variants: {
+      intent: makeVariants<'intent', ['error', 'info']>({
```
andresin87, line 10: intents might be:
- primary
- secondary
- success
- alert
- danger
- info
- neutral
- surface
for consistency

packages/components/badge/src/Badge.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Badge } from '.'
+
+type BadgeProps = ComponentProps<typeof Badge>
+
+const meta: Meta<typeof Badge> = {
+  title: 'Components/Badge',
+  component: Badge,
+}
+const sizes: BadgeProps['size'][] = ['sm', 'md']
+const intents: BadgeProps['intent'][] = ['error', 'info']
```
andresin87, line 16:
```suggestion
const intents: BadgeProps['intent'][] = ['primary', 'secondary', 'success', 'alert', 'danger', 'info', 'neutral', 'surface']
```

packages/components/badge/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@spark-ui/badge",
+  "version": "1.0.0",
```
andresin87, line 3:
```suggestion
  "version": "0.0.0",
```

packages/components/badge/src/BadgeItem.styles.tsx
andresin87: for deciding the color filling strategy: apply the same as button-filled design with no rules for hover-focus-select status

packages/components/badge/src/Badge.stories.tsx
kikoruiz: fixed ;)

PR #724, opened 2023-05-05
docs: add storylabel helper component

soykje:
To improve readability, we want to provide a shared `StoryLabel` component.

### Types of changes
- [x] 🧾 Documentation

### Screenshots - Animations
image


Powerplex:
Just something to keep in mind, all the markup markup in the stories is visible in the "show source" so it can add some noise, ex: 
Capture d’écran 2023-05-05 à 11 53 17

That's why on some components we directly used the native text of the component instead, ex: 
Capture d’écran 2023-05-05 à 11 54 07

Powerplex (approved):

Issue #734, opened 2023-05-08
[Components] Integrate FormField into Switch component


andresz1:
Enabling `FormField` for `Switch` component can be achieved by simply adding the `useFormFieldState()` hook. Consumer will then benefit from our field state components (error, hint, ...) :+1:

soykje:
For a11y concerns: https://www.w3.org/WAI/ARIA/apg/patterns/switch/

PR #782, opened 2023-05-16
feat(switch): allowing formfield usage for switch component

soykje:
**TASK**: #734 

We now want to enable the usage of `FormField` with our form components, to improve field state management and so.
- [x] adds `useFormFieldState()` hook within `Switch` component (it will default without throwing when no provider will be given)
- [x] adds dedicated tests, to improve confidence about the interaction between `Switch` and `FormField` (aria attributes completion, ...)
- [x] adds dedicated documentation section, with a link to `FormField` one for more details.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor

### Screenshots - Animations
image

### DIFF
--- a/packages/components/switch/src/Switch.doc.mdx
+++ b/packages/components/switch/src/Switch.doc.mdx
@@ -29,6 +29,12 @@ import { Switch } from "@spark-ui/switch"
 
 <Canvas of={stories.Default} />
 
+## Using a FormField
+
+If you need to provide more context about the Switch (for example, a helper message or an error message) the FormField component could be used. See [documentation](./?path=/docs/experimental-formfield--docs) for more details.
+
+<Canvas of={stories.ImprovedField} />
+
 ## Disabled
 
 <Canvas of={stories.Disabled} />
--- a/packages/components/switch/src/Switch.stories.tsx
+++ b/packages/components/switch/src/Switch.stories.tsx
@@ -1,5 +1,6 @@
 import { StoryLabel } from '@docs/helpers/StoryLabel'
 import { Button } from '@spark-ui/button'
+import { FormField } from '@spark-ui/form-field'
 import { EyeFill } from '@spark-ui/icons/dist/icons/EyeFill'
 import { EyeOffFill } from '@spark-ui/icons/dist/icons/EyeOffFill'
 import { Meta, StoryFn } from '@storybook/react'
@@ -26,6 +27,18 @@ export const Default: StoryFn = _args => (
       <Switch defaultChecked>Agreed</Switch>
     </div>
   </div>
+)
+
+export const ImprovedField: StoryFn = _args => (
+  <FormField name="agreement" isRequired>
+    <FormField.Label asChild>
+      <p>Agreement</p>
+    </FormField.Label>
+
+    <Switch>I agree</Switch>
+
+    <FormField.HelperMessage>Your agreement is important to us.</FormField.HelperMessage>
+  </FormField>
 )
 
 export const Disabled: StoryFn = _args => (
--- a/packages/components/switch/src/Switch.test.tsx
+++ b/packages/components/switch/src/Switch.test.tsx
@@ -1,4 +1,6 @@
+/* eslint-disable max-lines-per-function */
 /* eslint-disable max-nested-callbacks */
+import { FormField } from '@spark-ui/form-field'
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { useState } from 'react'
@@ -99,6 +101,51 @@ describe('Switch', () => {
 
       expect(spyOnCheckedChange).toHaveBeenCalledTimes(1)
       expect(screen.getByRole('switch')).toBeChecked()
+    })
+  })
+
+  describe('with FormField', () => {
+    it('should render with label and name', () => {
+      render(
+        <FormField name="agreement">
+          <FormField.Label asChild>
+            <p>Agreement</p>
+          </FormField.Label>
+
+          <Switch />
+        </FormField>
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toBeInTheDocument()
+    })
+
+    it('should render aria-attributes following FormField implementation', () => {
+      render(
+        <FormField name="agreement" isRequired isInvalid>
+          <FormField.Label asChild>
+            <p>Agreement</p>
+          </FormField.Label>
+
+          <Switch />
+
+          <FormField.ErrorMessage>Agreement is required</FormField.ErrorMessage>
+        </FormField>
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toHaveAttribute(
+        'aria-required',
+        'true'
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toHaveAttribute(
+        'aria-invalid',
+        'true'
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toHaveAttribute(
+        'aria-describedby',
+        screen.getByText('Agreement is required').getAttribute('id')
+      )
     })
   })
 })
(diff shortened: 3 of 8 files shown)


packages/components/switch/src/Switch.doc.mdx
```diff
@@ -29,6 +29,12 @@ import { Switch } from "@spark-ui/switch"
 
 <Canvas of={stories.Default} />
 
+## With FormField
+
+If you need more fine-grained control over the state of the Switch (requirement/validation feedbacks, ...) you can also use it within a `FormField` component. you can refer to [dedicated documentation](./?path=/docs/experimental-formfield--docs) for more details.
```
andresz1, line 34: If you need to provide more context about the switch (for example, a helper message or a error message) the `FormField` component could be used. See  [documentation](./?path=/docs/experimental-formfield--docs) for more details.

packages/components/switch/src/SwitchLabel.tsx
```diff
@@ -20,6 +21,14 @@ export interface LabelProps
   disabled?: boolean
 }
 
-export const Label = ({ className, disabled, ...others }: LabelProps) => {
-  return <RadixLabel className={labelStyles({ className, disabled })} {...others} />
+export const Label = ({ className, disabled, htmlFor, ...others }: LabelProps) => {
+  const { labelId } = useFormFieldState()
+
+  return (
+    <RadixLabel
```
andresz1, line 28: Here we should use spark label instead

packages/components/switch/src/SwitchInput.tsx
```diff
@@ -82,6 +95,13 @@ export const Input = React.forwardRef<HTMLButtonElement, InputProps>(
         checked={checked}
         defaultChecked={defaultChecked}
         onCheckedChange={handleCheckedChange}
+        id={id || controlledId}
+        name={name || controlledName}
+        required={required || isRequired}
+        aria-required={required || isRequired}
+        aria-invalid={isInvalid}
+        aria-labelledby={labelId}
```
andresz1, line 103: If there is already a label this is not needed 🤔 We can discuss it tomorrow if you want

packages/components/switch/src/Switch.doc.mdx
```diff
@@ -29,6 +29,12 @@ import { Switch } from "@spark-ui/switch"
 
 <Canvas of={stories.Default} />
 
+## With FormField
```
andresz1, line 32: `Using a form field` ?

packages/components/switch/src/Switch.doc.mdx
soykje: Done

packages/components/switch/src/SwitchInput.tsx
soykje: Based on the possibility to use Switch within a group, and/or following the GPDR settings pattern we saw together, I guess we should leave it as is, wdyt?

packages/components/switch/src/Switch.doc.mdx
soykje: Done

packages/components/switch/src/SwitchInput.tsx
soykje: Or should we omit this one for now, and wait for Design team to eventually add it?

packages/components/switch/src/Switch.test.tsx
```diff
@@ -101,4 +103,49 @@ describe('Switch', () => {
...
+        'aria-required',
+        'true'
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toHaveAttribute(
+        'aria-invalid',
+        'true'
+      )
+
+      expect(screen.getByRole('switch', { name: 'Agreement' })).toHaveAttribute(
```
andresz1, line 145: I just learnt this way of testing the description:

```
screen.getByRole('switch', { name: 'Agreement', description: 'Agreement is required' })
```

packages/components/switch/src/Switch.test.tsx
```diff
@@ -101,4 +103,49 @@ describe('Switch', () => {
...
+    it('should render aria-attributes following FormField implementation', () => {
+      render(
+        <FormField name="agreement" isRequired isInvalid>
+          <FormField.Label asChild>
+            <p>Agreement</p>
+          </FormField.Label>
+
+          <Switch />
+
+          <FormField.ErrorMessage>Agreement is required</FormField.ErrorMessage>
```
andresz1, line 131: Maybe would be nice to also test the helper message ?

packages/components/switch/src/SwitchInput.tsx
```diff
@@ -82,6 +95,13 @@ export const Input = React.forwardRef<HTMLButtonElement, InputProps>(
         checked={checked}
         defaultChecked={defaultChecked}
         onCheckedChange={handleCheckedChange}
+        id={id || controlledId}
+        name={name || controlledName}
+        required={required || isRequired}
+        aria-required={required || isRequired}
```
andresz1, line 101: There is not need of doing this. The spread operator will overwrite `{...rest}` these props

packages/components/switch/src/SwitchLabel.tsx
andresz1: Yes please! I think is better so we keep the label centralized in our codebase 🙏🏻

packages/components/switch/src/Switch.stories.tsx
```diff
@@ -28,6 +29,18 @@ export const Default: StoryFn = _args => (
   </div>
 )
 
+export const ImprovedField: StoryFn = _args => (
```
andresz1, line 32: I think that would be useful to have the following 4 examples:
- Invalid switch
- Switch with helper message
- Switch as required

Remember that if you add the `FormField.Label` the switch will have 2 labels

packages/components/switch/src/SwitchInput.tsx
andresz1: If there are `children` that means that a label is set so `aria-labelledby` is not needed.
```
aria-labelledby={children ? undefined : labelId}
```

andresz1: WDYT ?

andresz1: BTW I think that an story for a visually hidden label is needed but we can create another issue for that

packages/components/switch/src/SwitchInput.tsx
soykje: I'm not sure to understand :thinking: Even if the `required` attribute is correctly implemented with corresponding aria attribute, the `isRequired` fallback wouldn't be in there, don't you think? Or am I missing something?

packages/components/switch/src/SwitchInput.tsx
soykje: After doing some tests, it seems there is no accessible name available without adding this `aria-labelledby` :thinking: The `name` attribute doesn't seem to be added (Radix implementation) here, and the `role"switch"` doesn't match with the label (`Switch` child), as they're siblings.
I'm not sure if we should update the DOM structure, or if we should stick with this `aria-labelledby` anyway... wdyt?

packages/components/switch/src/Switch.test.tsx
soykje: I'm not sure of it. The FormField components (ErrorMessage, ...) should already be tested in related package. It would be duplicate to do it again here, don't you think?

packages/components/switch/src/Switch.stories.tsx
soykje: I'm not so sure, for the same reasons I'm not convinced we should add tests about FormField+Switch combo here. Those are not directly related to the component, but to an implementation.

Powerplex: I agree with @soykje , it's a "separation of concern" matter. 

I am already 50/50 about each component having a story to show how to combine it with `FormField`. So having every possible example I think will be too much to maintain whenever we update `FormField`. 

Regarding the "2 labels" thing maybe it can be an issue I am not sure, when I inspect the story it's true that the accessible name of the switch is no longer its label, but the formField label that replaces it. But maybe it's ok.
Capture d’écran 2023-05-25 à 15 29 40

andresz1 (approved):

acd02 (approved):

Issue #801, opened 2023-05-24
Checkbox icon not visible in Safari


andresz1:
### Description

When checkbox is checked, I expect the icon to be visible in Safari

### Steps to reproduce

1. Open Safari (desktop or mobile)
2. Go to checkbox documentation
3. Checkbox checked icon won't be visible

### Spark Component and Version

@spark-ui/checkbox@1.13.11

### Browser

Safari

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

PR #802, opened 2023-05-24
801 checkbox icon not visible in safari

andresz1:
## Checkbox icon not visible in safari

**TASK**: #801 

This PR fixes an issue where checked icon is not visible in Safari

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

Before
Screen Shot 2023-05-24 at 08 28 06

After
Screen Shot 2023-05-24 at 08 27 46

### DIFF
--- a/packages/components/checkbox/src/CheckboxIndicator.tsx
+++ b/packages/components/checkbox/src/CheckboxIndicator.tsx
@@ -10,7 +10,7 @@ export const CheckboxIndicator = forwardRef<HTMLSpanElement, CheckboxIndicatorPr
   (props, ref) => (
     <CheckboxIndicatorPrimitive
       ref={ref}
-      className="text-surface flex items-center justify-center"
+      className="text-surface flex h-full w-full items-center justify-center"
       {...props}
     />
   )
--- a/packages/components/checkbox/src/CheckboxInput.styles.ts
+++ b/packages/components/checkbox/src/CheckboxInput.styles.ts
@@ -3,10 +3,10 @@ import { cva, VariantProps } from 'class-variance-authority'
 
 export const checkboxInputStyles = cva(
   [
-    'h-sz-20 w-sz-20 border-md peer my-[var(--sz-1)] shrink-0 items-center justify-center self-baseline rounded-sm bg-transparent',
+    'h-sz-16 w-sz-16 border-md box-content items-center justify-center rounded-sm bg-transparent outline-none',
     'spark-disabled:opacity-dim-3 spark-disabled:cursor-not-allowed spark-disabled:hover:ring-0',
-    'focus-visible:ring-outline-high focus-visible:outline-none focus-visible:ring-2',
-    'hover:border-primary-container hover:outline-none hover:ring-2',
+    'focus-visible:ring-outline-high focus-visible:ring-2',
+    'hover:border-primary-container hover:ring-2',
     'u-shadow-border-transition',
   ],
   {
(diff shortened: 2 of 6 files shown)


packages/components/checkbox/src/CheckboxInput.styles.ts
```diff
@@ -3,10 +3,10 @@ import { cva, VariantProps } from 'class-variance-authority'
 
 export const checkboxInputStyles = cva(
   [
-    'h-sz-20 w-sz-20 border-md peer my-[var(--sz-1)] shrink-0 items-center justify-center self-baseline rounded-sm bg-transparent',
```
andresz1, line 6: @kikoruiz I removed some stuff that I thought it wasn't necessary. Can you double check please ? 🙏🏻

packages/components/checkbox/src/CheckboxInput.tsx
```diff
@@ -62,7 +63,7 @@ export const CheckboxInput = forwardRef<HTMLButtonElement, CheckboxInputProps>(
         {...others}
       >
         <CheckboxIndicator>
-          {checked === 'indeterminate' ? indeterminateIcon : icon}
+          <Icon size="sm">{checked === 'indeterminate' ? indeterminateIcon : icon}</Icon>
```
andresz1, line 66: Not 100% sure if it's ok to use icon size here 🤔 WDYT ?

packages/components/checkbox/src/CheckboxInput.styles.ts
kikoruiz: I did't work on this component before but I guess the `peer` classname is not needed by any other sibling element, right?

kikoruiz (approved):

packages/components/checkbox/src/CheckboxIndicator.tsx
```diff
@@ -10,7 +10,7 @@ export const CheckboxIndicator = forwardRef<HTMLSpanElement, CheckboxIndicatorPr
   (props, ref) => (
     <CheckboxIndicatorPrimitive
       ref={ref}
-      className="text-surface flex items-center justify-center"
+      className="text-surface flex h-full w-full items-center justify-center"
```
andresin87, line 13: i think this classes should be placed in xxxxx.styles.tsx to decouple the view from the logic

andresin87 (approved):

packages/components/checkbox/src/CheckboxIndicator.tsx
Powerplex: I think it is one of the strengths of Tailwind to be able to apply some styles on the markup. It might not be necessary to always split the two ?

packages/components/checkbox/src/CheckboxInput.styles.ts
Powerplex: I checked on the main branch and `peer` was no longer needed :)

packages/components/checkbox/src/CheckboxInput.tsx
Powerplex: I think it's fine because checkbox have a unique size :)

Powerplex (approved):

packages/components/checkbox/src/CheckboxIndicator.tsx
andresz1: I agree with @Powerplex but we can discuss it later on

Issue #796, opened 2023-05-23
Improve accessibility on Switch component


soykje:
### Description

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself, identified through the `role="switch"` attribute, doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an `htmlFor` attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case (if Switch doesn't have any children...)

### Problem Statement/Justification

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself, identified through the `role="switch"` attribute, doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

![image](https://github.com/adevinta/spark/assets/66770550/fbe178d2-8dcc-43f4-8913-031e0d228c92)

### Proposed Solution or API

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an `htmlFor` attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case (if Switch doesn't have any children...)

```tsx
import React from 'react';
import * as Switch from '@radix-ui/react-switch';

const SwitchDemo = () => (
  <form>
    <div>
      <label htmlFor="airplane-mode">
        Airplane mode
      </label>
      <Switch.Root id="airplane-mode">
        <Switch.Thumb />
      </Switch.Root>
    </div>
  </form>
);

export default SwitchDemo;
```

PR #809, opened 2023-05-24
refactor(switch): fix and improve accessibility on switch component

soykje:
**TASK**: #796 

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself, identified through the `role="switch"` attribute, doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an `htmlFor` attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case (if Switch doesn't have any children...)
- 
### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
**Before:**
image

**After:**
image


packages/components/switch/src/Switch.doc.mdx
```diff
@@ -62,6 +62,8 @@ It is then up to you to trigger update to its `checked` state to toggle it.
 
 ## Accessibility
 
+This component should always have a label, directly declared through the `children` prop or inherited from the `FormField` wrapper. If none of those are available it is strongly recommended to define an `aria-label` attribute, to comply to accessibility requirements.
```
andresz1, line 65: 👏🏻

packages/components/switch/package.json
```diff
@@ -17,6 +17,7 @@
     "directory": "packages/components/switch"
   },
   "dependencies": {
+    "@radix-ui/react-id": "^1.0.0",
```
andresz1, line 20: There's an agreement of using fixed version for external dependencies

packages/components/switch/src/Switch.tsx
```diff
@@ -6,12 +8,45 @@ import { Label } from './SwitchLabel'
...
+        <Input
+          ref={ref}
+          size={size}
+          id={controlledInputId || switchId}
+          disabled={disabled}
+          /**
+           * If the switch doesn't have any direct label (children) then we should try to
+           * get an eventual alternative label from FormField.
+           * On last resort, we shouldn't forget to define an aria-label attribute.
+           */
```
andresz1, line 28: IMO tests should explain this part

packages/components/switch/src/Switch.tsx
```diff
@@ -6,12 +8,45 @@ import { Label } from './SwitchLabel'
...
+          {...(children
+            ? {
+                'aria-labelledby': switchLabelId,
+              }
+            : {
+                'aria-labelledby': controlledLabelId,
+              })}
```
andresz1, line 35: Why do you need to create an object here? My proposal:
```jsx
aria-labelledby={children ? switchLabelId : controlledLabelId}
```

packages/components/switch/src/Switch.tsx
```diff
@@ -6,12 +8,45 @@ import { Label } from './SwitchLabel'
 export type SwitchProps = InputProps
 
 export const Switch = forwardRef<HTMLButtonElement, SwitchProps>(
-  ({ value = 'on', size = 'md', children, className, ...rest }, ref) => {
+  ({ value = 'on', size = 'md', children, className, id, disabled, ...rest }, ref) => {
+    const { id: controlledInputId, labelId: controlledLabelId } = useFormFieldState()
+
+    const switchId = useId(id)
```
andresz1, line 14: IMO `switch` prefix is not needed it's redundant

packages/components/switch/src/Switch.tsx
```diff
@@ -6,12 +8,45 @@ import { Label } from './SwitchLabel'
 export type SwitchProps = InputProps
 
 export const Switch = forwardRef<HTMLButtonElement, SwitchProps>(
-  ({ value = 'on', size = 'md', children, className, ...rest }, ref) => {
+  ({ value = 'on', size = 'md', children, className, id, disabled, ...rest }, ref) => {
+    const { id: controlledInputId, labelId: controlledLabelId } = useFormFieldState()
```
andresz1, line 12: The prefix controlled could be confusing I think we should use the word `field` somehow. If you don't destructure the object I think it would be better. For example:
```jsx
const field = useFormFieldState()
console.log(field.id)
```

packages/components/switch/src/Switch.tsx
soykje: I understand your point. Wdyt of `innerId`, for example?

packages/components/switch/src/Switch.tsx
soykje: Wdyt? https://github.com/adevinta/spark/commit/576488518abba5e747a31386ac92ce9a97d3da0a

packages/components/switch/package.json
soykje: https://github.com/adevinta/spark/wiki/Agreements#dependencies-versionning-on-packagejson

packages/components/switch/src/Switch.stories.tsx
```diff
@@ -85,9 +85,9 @@ const switchColors: ComponentProps<typeof Switch>['intent'][] = [
 export const Colors: StoryFn = _args => (
   <div className="gap-lg flex">
     {switchColors.map(color => (
-      <div key={color}>
-        <StoryLabel>{color}</StoryLabel>
-        <Switch name="primary" intent={color} defaultChecked />
+      <div key={String(color)}>
```
acd02, line 88: Why do you have to stringify `color`?

acd02 (approved):

PR #814, opened 2023-05-25
refactor: widen the Tailwind peerDep for utils

DanArthurGallagher:

The Utility package for Tailwind has a pinned dependency on the exact major of TailwindCSS, which is restrictive for a consumer. If there is no limiting issue, and given how liberally SemVer is used in the JS world, I propose to pin only to the major and allow consumers to upgrade.


ale-grosselle (approved):

andresz1 (approved):

andresin87 (approved):

kikoruiz (approved):

andresz1:
LGTM thanks @DanArthurGallagher DanArthurGallagher

PR #816, opened 2023-05-25
feat(use-combined-state): remove unnecessary state management when controlled state mode

andresin87:
**TASK**: 

reduces the rerenders and the complexity.

### Types of changes

- [x] 🧠 Refactor

https://github.com/chakra-ui/chakra-ui/blob/main/packages/legacy/hooks/src/use-controllable.ts#L34


packages/hooks/use-combined-state/src/useCombinedState.tsx
```diff
@@ -1,34 +1,38 @@
...
 
 /**
- * This hook must be used when a component have both a controlled and uncontrolled mode.
- * It will take care of updating the state value when controlled mode (prop) is updated.
+ * This hook must be used when a component has both a controlled and uncontrolled mode.
+ * It will take care of updating the state value when a controlled mode (prop) is updated.
  */
 export function useCombinedState<T>(
   controlledValue?: T,
   defaultValue?: T
```
andresz1, line 10: Great job @andresin87 do you think that would be possible to add the `onChange` callback like chakra does ? 🙏🏻 I think that could be really useful

andresin87: ok. I'll do today

andresz1 (approved):

packages/hooks/use-combined-state/src/useCombinedState.tsx
andresin87: 3rd argument

packages/hooks/use-combined-state/src/useCombinedState.test.tsx
```diff
@@ -131,4 +131,62 @@ describe('useCombinedState', () => {
...
+  it('value must NOT be updated using its setter when using uncontrolled mode using shouldUpdate flag to false', () => {
+    // Given
+    const args = { value: undefined, defaultValue: 123 } as argsType
+
+    // When
+    const hook = renderHook(() => useCombinedState(...map(args)))
+    let [response, setResponse, isControlledValue, initialValue] = hook.result.current
+
+    // Then
+    expect(typeof response).toBe('number')
```
Powerplex, line 144: I think you might not need some assertions like this one :) We have type safety using TS + other checks implicitely covering it:
`expect(response).toBe(123)`

Powerplex (approved):

packages/hooks/use-combined-state/src/useCombinedState.test.tsx
andresin87: DONE!

packages/hooks/use-combined-state/src/useCombinedState.tsx
andresin87: DONE!

kikoruiz (approved):

Issue #799, opened 2023-05-23
Improve accessibility on Checkbox component


soykje:
### Description

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an htmlFor attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case

### Problem Statement/Justification

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

![image](https://github.com/adevinta/spark/assets/66770550/07da177d-ee3b-4bb4-893f-10efaf1d64c1)

### Proposed Solution or API

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an htmlFor attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case

```tsx
import React from 'react';
import * as Switch from '@radix-ui/react-switch';
import './styles.css';

const SwitchDemo = () => (
  <form>
    <div style={{ display: 'flex', alignItems: 'center' }}>
      <label className="Label" htmlFor="airplane-mode" style={{ paddingRight: 15 }}>
        Airplane mode
      </label>
      <Switch.Root className="SwitchRoot" id="airplane-mode">
        <Switch.Thumb className="SwitchThumb" />
      </Switch.Root>
    </div>
  </form>
);

export default SwitchDemo;
```

PR #832, opened 2023-05-26
fix(checkbox): fix and improve a11y on checkbox

soykje:
**TASK**: #799 

With current DOM structure (label tag wrapping the whole content) we recently noticed a small accessibility concern, about accessible name.

Indeed, the component itself doesn't have any accessible name. This is due to component text (aka children) added as a sibling DOM node, and so not directly related. This can be confirmed using the "Axe" Chrome accessibility extension, and testing the component.

After discussions with @andresz1, @acd02 and @Powerplex we agreed:
- changing the DOM structure, to match the Radix pattern: having the label as a sibling, linked through an htmlFor attribute. The whole content wrapped within a neutral div, for ex.
- using `aria-label` or `aria-labelledby` based on use case

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor

### Screenshots - Animations
**Before:**
image

**After:**
image


packages/components/checkbox/src/Checkbox.test.tsx
```diff
@@ -195,115 +194,52 @@ describe('Checkbox', () => {
     expect(onCheckedChange).toHaveBeenCalledWith(['baseball', 'soccer'])
   })
 
-  it('should render group with a label', () => {
```
andresz1, line 198: I don't understand why all these tests are being modified/removed

andresz1 (changes requested):

packages/components/checkbox/src/Checkbox.test.tsx
```diff
@@ -195,115 +194,52 @@ describe('Checkbox', () => {
...
+      )
 
-    const groupEl = screen.getByRole('group', {
-      name: 'Sports',
-      description: 'Sports combination is invalid',
+      expect(screen.getByRole('group', { name: 'Sports' })).toBeInTheDocument()
     })
 
-    expect(groupEl).toBeInTheDocument()
+    it('should render aria-attributes following FormField implementation', () => {
```
andresz1, line 213: - The more our tests resemble the way our users will interact with our components the better.
- I do not see the need of coupling this test with `FormField` concept.
- IMO this tests is too generic

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -20,41 +21,50 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
       disabled,
       onCheckedChange,
       children,
-      ...others
+      ...rest
     },
     forwardedRef
   ) => {
+    const checkboxId = useId()
```
andresz1, line 28: IMO there is no need of using `checkbox` prefix is redundant since the component is named `Checkbox`

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -20,41 +21,50 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
...
+    const checkboxLabelId = useId()
```
andresz1, line 29: Same as above

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -20,41 +21,50 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
       disabled,
       onCheckedChange,
       children,
-      ...others
+      ...rest
```
andresz1, line 24: We don't have an agreement regarding this name. Maybe should come with agreement ? In most of our components is named `others`

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -64,10 +74,27 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
           aria-describedby={description}
           aria-invalid={isInvalid}
           onCheckedChange={handleCheckedChange}
-          {...others}
+          /**
+           * If the checkbox doesn't have any direct label (children) then we should try to
+           * get an eventual alternative label from FormField.
+           * On last resort, we shouldn't forget to define an aria-label attribute.
+           */
+          {...(children
```
andresz1, line 82: There is no need of using the spread operator here.
```jsx
aria-labelledby={children ? labelId : field.labelId}
```

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -64,10 +74,27 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
           aria-describedby={description}
           aria-invalid={isInvalid}
           onCheckedChange={handleCheckedChange}
-          {...others}
+          /**
+           * If the checkbox doesn't have any direct label (children) then we should try to
+           * get an eventual alternative label from FormField.
+           * On last resort, we shouldn't forget to define an aria-label attribute.
+           */
```
andresz1, line 81: IMO no need for this comment should be tested and explained in the tests

packages/components/checkbox/src/CheckboxLabel.tsx
```diff
@@ -9,8 +9,8 @@ export interface CheckboxLabelProps extends LabelProps, LabelStylesProps {
   disabled?: boolean
 }
 
-export const CheckboxLabel = ({ className, disabled, ...others }: CheckboxLabelProps) => {
-  return <Label className={labelStyles({ className, disabled })} {...others} />
```
andresz1, line 13: Why are you removing the `className` here ? 🤔

soykje: The `className` was previously provided to `Label` as it was wrapping the whole component. Now it has been changed, so the prop will remain on the main wrapper, not here anymore.

packages/components/checkbox/src/Checkbox.tsx
soykje: Well, I think, just as `Switch`, that both could be useful. Wdyt? https://github.com/adevinta/spark/pull/832/commits/06e5a6d2af6bdc42149c1042cbedcfae2a30fee3

andresin87 (approved):

packages/components/checkbox/src/Checkbox.tsx
andresin87: We always need an Id for each element/component?

soykje: This is just for this one, that needs an accessible name. The usecase shouldn't occur here, but it's technically possible, just as for the Switch component. So for the sake of consistency I choosed to use same approach.

packages/components/checkbox/src/Checkbox.test.tsx
```diff
@@ -195,115 +194,87 @@ describe('Checkbox', () => {
...
-        <CheckboxGroup>
-          <Checkbox value="soccer">Soccer</Checkbox>
-          <Checkbox value="baseball">Baseball</Checkbox>
-        </CheckboxGroup>
-
-        <FormField.ErrorMessage>Sports combination is invalid</FormField.ErrorMessage>
-      </FormField>
-    )
+          <CheckboxGroup>
+            <Checkbox value="soccer">Soccer</Checkbox>
```
kikoruiz, line 239: Soccer? Really? 😜

kikoruiz (approved):

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -9,6 +11,7 @@ import { CheckboxLabel } from './CheckboxLabel'
 export type CheckboxProps = CheckboxInputProps
 
 export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
+  // eslint-disable-next-line complexity
```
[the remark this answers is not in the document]
acd02: It's because of the `intent` const, replace with what we currently have and you won't need this eslint comment:
```tsx
const intent = (() => {
  if (isInvalid) return 'error'

  return intentProp ?? group.intent
})()
```

Or, you could make it a function:
```tsx
const getIntent = () => (isInvalid ? 'error' : intentProp ?? group.intent)
```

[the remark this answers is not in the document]
soykje: Yes I noticed that :+1: But imho this syntax is less readable, and this is why I was looking for a better solution... But If nothing pops, I'll go back with your method :+1:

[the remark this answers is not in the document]
soykje: @acd02 Just updated the code...wdyt? I guess it's mainly about personal preferences, but at least there is no more `eslint-disable-next-line complexity` :sweat_smile:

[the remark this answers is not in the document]
acd02: Not really fond of the names, `getFieldOrGroupAttribute`, `getFieldOrGroupAttribute` 😄,
Why not go all the way and make something like:
```tsx
interface GetFieldAttributes {
  field: ReturnType<typeof useFormFieldState>
  group: ReturnType<typeof useCheckboxGroup>
  intent: CheckboxInputProps['intent']
}

function getFieldAttributes({ field, group, intent: intentProp }: GetFieldAttributes) {
  const name = field.name ?? group.name
  const isRequired = field.isRequired ?? group.isRequired
  const isInvalid = field.isInvalid ?? group.isInvalid
  const isFieldEnclosed = field.id !== group.id
  const id = isFieldEnclosed ? field.id : undefined
  const description = isFieldEnclosed ? field.description : undefined
  const intent = isInvalid ? 'error' : intentProp ?? group.intent

  return { name, isRequired, isInvalid, isFieldEnclosed, id, description, intent }
}

...

const { id, name, isInvalid, isRequired, description, intent } = getFieldAttributes({
  field,
  group,
  intent: intentProp,
})

...

return (
      <CheckboxLabel>
        <CheckboxInput
          ref={ref}
          id={id}
          name={name}
          value={value}
          intent={intent}
          checked={checked}
          disabled={disabled}
          required={isRequired}
          aria-describedby={description}
          aria-invalid={isInvalid}
          onCheckedChange={handleCheckedChange}
          {...others}
        />
        {children}
      </CheckboxLabel>
    )

```

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -22,41 +24,60 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
...
     const rootRef = useRef<HTMLButtonElement | undefined>()
     const ref = useMergeRefs(forwardedRef, rootRef)
 
-    const name = field.name ?? group.name
-    const isRequired = field.isRequired ?? group.isRequired
-    const isInvalid = field.isInvalid ?? group.isInvalid
-    const isFieldEnclosed = field.id !== group.id
-    const id = isFieldEnclosed ? field.id : undefined
-    const description = isFieldEnclosed ? field.description : undefined
+    const getCheckboxAttributes = ({
```
acd02, line 36: defining this function outside of the `Checkbox` component would simplify it even further imho

andresz1:
There is a difference to what we had before and is that now is not clickable in between the checkbox and the label. Is this something intended ?

https://github.com/adevinta/spark/assets/6877967/a0326098-bc28-4beb-a485-93de8114c095

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -22,41 +24,38 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
...
     const handleCheckedChange = (isChecked: boolean) => {
-      if (onCheckedChange) onCheckedChange(isChecked)
+      onCheckedChange?.(isChecked)
 
       const rootRefValue = rootRef.current?.value
-
-      if (!rootRefValue || !group.onCheckedChange) return
-
-      group.onCheckedChange(isChecked, rootRefValue)
+      rootRefValue && group.onCheckedChange?.(isChecked, rootRefValue)
```
andresz1, line 42: I think the other way around was more legible. As a general rule I never inline `if` statements like you are doing

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -66,12 +65,45 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
           aria-describedby={description}
           aria-invalid={isInvalid}
           onCheckedChange={handleCheckedChange}
+          /**
+           * If the checkbox doesn't have any direct label (children) then we should try to
+           * get an eventual alternative label from FormField.
+           * On last resort, we shouldn't forget to define an aria-label attribute.
+           */
+          aria-labelledby={children ? innerLabelId : field.labelId}
```
andresz1, line 73: This is not needed if you are already using `htmlFor`

andresz1 (approved):

packages/components/checkbox/src/Checkbox.test.tsx
andresz1: I don't get why this refactor is needed (tests and documentation) maybe I'm missing something but the issue was to improve the a11y of the component and most of the changes aren't related with it.

packages/components/checkbox/src/CheckboxLabel.tsx
andresz1: Ok I see. However, as an agreement we said that all components should extend the `className`

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -66,12 +65,45 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
...
+          <CheckboxLabel disabled={disabled} htmlFor={id || innerId} id={innerLabelId}>
+            {children}
+          </CheckboxLabel>
+        )}
+      </div>
     )
   }
 )
 
+const getCheckboxAttributes = ({
```
andresz1, line 87: I'm not sure about doing this. What's the reasoning of creating functions outside components when they are no being reused somewhere else ?

acd02: To provide more context, the code had been refactored due to an ESLint warning. The function's complexity exceeded the threshold of the ESLint rule.

And I was the one suggesting @soykje to define the function outside of the component (for the rationale behind this suggestion, see my comment just above)

soykje: Yes, your fault! :stuck_out_tongue_closed_eyes: Seriously, I'm not a big fan of this syntax too, but as Alex said, this is mainly ESLint faul... we have to respect those rules, or discuss on how we make them evolve :slightly_smiling_face:

packages/components/checkbox/src/Checkbox.test.tsx
soykje: Some of the changes here are related to how we documented and tested `FormField` usage. As discussed with you and Julien (if I remember correctly) we agreed to remove sections (within tests and documentation) that were too closely related to `FormField` implementation.
It could have been made within another PR I guess, if it's your point, but maybe now, after two weeks, we can leave it as is? :innocent:

andresz1: 👌🏻

packages/components/checkbox/src/CheckboxLabel.tsx
soykje: Oh, didn't know. But does this agreement concern also "subcomponents"? As this is not directy exposed and do not have any usage of this prop, it doesn't make really sense to me... don't you think? :thinking:

soykje:
> 

Hmmm... Well not intended no :disappointed: I tested that the label was still clickable, but didn't noticed about this space between both elements. Couldn't say, if it is an issue or not, but it is the same behaviour we have for radio and switch, if I'm not wrong... So maybe it's ok?

packages/components/checkbox/src/CheckboxLabel.tsx
andresin87: any sub-component exposed (exported) must follow the same agreements

packages/components/checkbox/src/Checkbox.tsx
soykje: As discussed, we'll move this function back within `Checkbox`. We'll discuss later about how we live with ESLint :sunglasses:

acd02: 🥲

packages/components/checkbox/src/Checkbox.tsx
soykje: I'm not so sure... Maybe it is related to our checkbox element being a button and not an input, but the `htmlFor` doesn't seem to be enough:
image

andresz1 (approved):

PR #847, opened 2023-05-31
[CHORE] fix storybook build

kikoruiz:
### Types of changes


packages/components/popover/package.json
```diff
@@ -22,6 +22,6 @@
     "directory": "packages/components/popover"
   },
   "dependencies": {
-    "@radix-ui/react-popover": "1.0.6-rc.13"
+    "@radix-ui/react-popover": "1.0.6"
```
soykje, line 25: I'm always lost with version naming... Just to be sure: `1.0.6` > `1.0.6-rc.13`? Because the RC was specifically related to a bugfix https://github.com/radix-ui/primitives/issues/2128#issuecomment-1557314630

soykje, line 25: :+1:

package.json
```diff
@@ -6,6 +6,7 @@
   "license": "MIT",
   "scripts": {
     "prepare": "husky install",
+    "install:lock": "npm install --package-lock-only --legacy-peer-deps --no-audit --no-fund",
```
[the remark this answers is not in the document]
kikoruiz: To manually fix the lockfile from your local if you need

soykje (approved):

packages/components/popover/package.json
kikoruiz: The final release was done some time after the Popover was merged to master.

package.json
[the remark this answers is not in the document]
kikoruiz: Yep, we need to agree on that as well 😓, I'm removing it from now, it's not so important

andresz1 (approved):

Issue #844, opened 2023-05-30
Remove ESLint ignore comments


acd02:
In our codebase, there are several ESLint ignore comments that have been added to disable specific rules or warnings. These comments can make it harder to maintain and understand the codebase over time.

We should remove these ignore comments and instead address the underlying issues they are suppressing.

PR #848, opened 2023-05-31
remove eslint ignore comments

acd02:
**TASK**: #844

In our codebase, there are several ESLint ignore comments that have been added to disable specific rules or warnings. These comments can make it harder to maintain and understand the codebase over time.

This PR remove these comments and instead address the underlying issues they are suppressing.

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [x] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/tabs/src/TabsList.tsx
+++ b/packages/components/tabs/src/TabsList.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-lines-per-function */
 import * as RadixTabs from '@radix-ui/react-tabs'
 import { Button } from '@spark-ui/button'
 import { Icon } from '@spark-ui/icon'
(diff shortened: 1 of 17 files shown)


packages/components/tabs/src/TabsList.tsx
```diff
@@ -134,19 +133,7 @@ export const TabsList = forwardRef<HTMLDivElement, TabsListProps>(
...
-            className={navigationArrowStyles()}
-            onClick={handlePrevClick}
-            disabled={arrows.prev === 'disabled'}
-            aria-label="Scroll left"
-          >
-            <Icon>
-              <ArrowVerticalLeft />
-            </Icon>
-          </Button>
+          <Components.Prev handleClick={handlePrevClick} disabled={arrows.prev === 'disabled'} />
```
soykje, line 136: I personnally don't like this way to write code... Also if the only purpose of this change is to comply to ESlint rule, maybe we should change/disable it, wdyt? :fist_raised:

acd02: We can disable the rule for this file, but imho, we should keep this rule. Getting a warning when a function exceeds 100 lines serves as a good guardrail, indicating when we should consider splitting a function into smaller pieces.

acd02: I wish we had a way to differentiate normal functions from component functions, which contain markup and can quickly grow in number of lines

acd02: Found a middle ground and increased the threshold for `tsx` files to 150 lines. 👍

soykje: I get the point, but then maybe we could simply put those buttons components code in separate file(s), wdyt? They shouldn't be exported at package level but internally it seems it could be a good middle ground?

soykje (approved):

PR #870, opened 2023-06-05
test(popover): set of tests for Popover package

Powerplex:
**TASK**: #869 

Experimental Popover package was missing unit tests.

### Types of changes
- [x] 🧪 Test

### DIFF
--- a/packages/components/popover/package.json
+++ b/packages/components/popover/package.json
@@ -23,5 +23,8 @@
   },
   "dependencies": {
     "@radix-ui/react-popover": "1.0.6"
+  },
+  "devDependencies": {
+    "jsdom-testing-mocks": "1.9.0"
   }
 }
--- a/packages/components/popover/src/Popover.test.tsx
+++ b/packages/components/popover/src/Popover.test.tsx
@@ -1,27 +1,322 @@
+/* eslint-disable max-nested-callbacks */
+/* eslint-disable max-lines-per-function */
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
+import { mockResizeObserver } from 'jsdom-testing-mocks'
+import { useState } from 'react'
 import { describe, expect, it, vi } from 'vitest'
 
 import { Popover } from './Popover'
 
 describe('Popover', () => {
-  it('should render', () => {
-    render(<Popover>Hello World!</Popover>)
-
-    expect(screen.getByText('Hello World!')).toBeInTheDocument()
+  beforeAll(() => {
+    mockResizeObserver()
   })
 
-  it('should trigger click event', async () => {
-    const user = userEvent.setup()
-    const clickEvent = vi.fn()
+  describe('uncontrolled mode (stateful)', () => {
+    describe('opening', () => {
+      it('should open popover on click', async () => {
+        const user = userEvent.setup()
 
-    // Given
-    render(<div onClick={clickEvent}>Hello World!</div>)
+        // Given a basic popover (closed by default)
+        render(
+          <Popover>
+            <Popover.Trigger asChild>
+              <button type="button">Click me</button>
+            </Popover.Trigger>
+            <Popover.Anchor>
+              <button type="button">Popover anchor</button>
+            </Popover.Anchor>
+            <Popover.Content>
+              <Popover.Arrow />
+              Popover content
+            </Popover.Content>
+          </Popover>
+        )
 
-    // When
-    await user.click(screen.getByText('Hello World!'))
+        // ...by default, only trigger elements are rendered
+        expect(
+          document.querySelector('[data-spark-component="popover-anchor"]')
+        ).toBeInTheDocument()
+        expect(
+          document.querySelector('[data-spark-component="popover-trigger"]')
+        ).toBeInTheDocument()
 
-    // Then
-    expect(clickEvent).toHaveBeenCalledTimes(1)
+        // ...and content elements are not rendered
+        expect(document.querySelector('[data-spark-component="popover-arrow"]')).toBeNull()
+        expect(document.querySelector('[data-spark-component="popover-content"]')).toBeNull()
+        expect(screen.queryByText('Popover content')).not.toBeInTheDocument()
+
+        // When the user click on its associated trigger element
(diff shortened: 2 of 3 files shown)


packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,309 @@
...
+              <Popover.Content onOpenAutoFocus={focusEvent}>
+                Popover content
+                <input type="text" placeholder="First focusable element" />
+              </Popover.Content>
+            </Popover>
+          )
+
+          // Then the first focusable child does not get the focus because the event handler cancels it
+          expect(focusEvent).toHaveBeenCalledTimes(1)
+          expect(document.activeElement).not.toBe(
```
soykje, line 113: I'm wondering if we couldn't simply use something like:
```tsx
expect(screen.getByPlaceholderText('First focusable element')).toHaveFocus()
```
?

packages/components/popover/package.json
```diff
@@ -23,5 +23,8 @@
   },
   "dependencies": {
     "@radix-ui/react-popover": "1.0.6"
+  },
+  "devDependencies": {
+    "jsdom-testing-mocks": "^1.9.0"
```
soykje, line 28:
```suggestion
    "jsdom-testing-mocks": "1.9.0"
```
Please see our last [agreements](https://github.com/adevinta/spark/wiki/Agreements#managing-dependencies-within-spark-components) :+1:

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
+import { useState } from 'react'
 import { describe, expect, it, vi } from 'vitest'
 
 import { Popover } from './Popover'
 
+const getSparkComponent = (componentName: string): Element | null => {
+  return document.querySelector(`[data-spark-component="${componentName}"]`)
+}
+
+mockResizeObserver()
```
soykje, line 15: Why not within a `beforeAll` block?

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
 import userEvent from '@testing-library/user-event'
+import { mockResizeObserver } from 'jsdom-testing-mocks'
+import { useState } from 'react'
 import { describe, expect, it, vi } from 'vitest'
 
 import { Popover } from './Popover'
 
+const getSparkComponent = (componentName: string): Element | null => {
+  return document.querySelector(`[data-spark-component="${componentName}"]`)
+}
```
andresz1, line 13: IMO for tests I prefer to repeat the code so I don't need to waste time looking for functions

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
+        expect(getSparkComponent('popover-anchor')).toBeInTheDocument()
+        expect(getSparkComponent('popover-trigger')).toBeInTheDocument()
+
+        // ...and content elements are not rendered
+        expect(getSparkComponent('popover-arrow')).toBeNull()
+        expect(getSparkComponent('popover-content')).toBeNull()
+        expect(screen.queryByText('Popover content')).not.toBeInTheDocument()
+
+        // When the user click on its associated trigger element
+        await user.click(screen.getByText('Click me'))
```
andresz1, line 49:
```suggestion
        await user.click(screen.getByRole('button', { name: 'Click me' }))
```

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
+        it('should forward focus to first focusable child', () => {
+          // Given a onOpenAutoFocus event
+          const focusEvent = vi.fn()
+
+          // When an opened popover is rendered
+          render(
+            <Popover defaultOpen>
+              <Popover.Trigger asChild>
+                <button type="button">Click me</button>
+              </Popover.Trigger>
```
andresz1, line 82: Is the `type` needed here ?

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
+          )
+
+          // Then first focusable child has focus and event has been called
+          expect(focusEvent).toHaveBeenCalledTimes(1)
+          expect(screen.getByPlaceholderText('First focusable element')).toHaveFocus()
+        })
+
+        it('should not forward focus to first focusable child when using preventDefault', () => {
+          // Given a onOpenAutoFocus event preventing event from being executed
+          const focusEvent = vi.fn(e => {
```
andresz1, line 97:
```suggestion
          const onOpenAutoFocus = vi.fn(e => {
```

packages/components/popover/src/Popover.test.tsx
```diff
@@ -1,27 +1,318 @@
...
+              </Popover.Trigger>
+              <Popover.Content onCloseAutoFocus={onCloseAutoFocus}>
+                Popover content
+                <input type="text" placeholder="First focusable element" />
+              </Popover.Content>
+            </Popover>
+          )
+
+          // When the user open the popover
+          await user.click(screen.getByText('Click me'))
```
andresz1, line 245: Better to `getByRole`

andresz1 (approved):

packages/components/popover/src/Popover.test.tsx
Powerplex: Yes :) I use `getByRole` only when I want to make sure of the element type. But in that case the button is not part of the Popover component I don't need to check its role (because getByRole is slower and the cause of performance issue on Leboncoin)

packages/components/popover/src/Popover.test.tsx
Powerplex: I remember that button should always have an explicit type :) https://dev.to/clairecodes/why-its-important-to-give-your-html-button-a-type-58k9

packages/components/popover/src/Popover.test.tsx
andresz1: For me is more related with readability purposes in this case but I get your point. Do you have any source related to `getByRole` being slower ?

PR #877, opened 2023-06-07
feat(chip)

andresin87:
**TASK**: 

### Types of changes


Powerplex:
Maybe we should remove the `icon` prop:
Capture d’écran 2023-06-13 à 15 31 44

And follow the same logic as in other components (composability using `children` prop):
Capture d’écran 2023-06-09 à 16 50 47

WDYT @andresin87 ?

andresin87:
> Maybe we should remove the `icon` prop: Capture d’écran 2023-06-13 à 15 31 44
> 
> And follow the same logic as in other components (composability using `children` prop):
> 
> Capture d’écran 2023-06-13 à 15 31 00
> 
> WDYT @andresin87 ?

how do we apply the labels to that icons if we do that? `icon-label`prop?

Powerplex:
@andresin87 You mean to apply an aria-label to an icon that should not be decorative only ? `Icon` has a `label ` prop for that.

I think both approach are the some regarding applying an aria-label to the icon, no ?

packages/components/chip/src/Chip.styles.tsx
```diff
@@ -0,0 +1,183 @@
...
+    },
+  }
+)
+
+export const chipContentStyles = cva(
+  ['gap-sm inline-flex h-full items-center justify-center whitespace-nowrap'],
+  {
+    variants: {
+      isBordered: {
+        no: ['max-w-[240px]'],
```
Powerplex, line 73: `max-w-sz-240`. (but for the 238px you must keep the arbitrary value :) )

packages/components/chip/src/Chip.styles.tsx
```diff
@@ -0,0 +1,183 @@
...
+        mode: 'icon',
+        hasCloseIcon: 'yes',
+        isBordered: 'yes',
+        class: tw(['pl-md']),
+      },
+    ],
+  }
+)
+
+export const chipContentTextStyles = cva(['overflow-hidden text-ellipsis'])
```
Powerplex, line 143: Imho this is the type of things where using `cva` and calling its function is overkill, and we should simply use tailwind classes on the markup.

Or at least we could use `cx()` instead of `cva()` (which is meant to manage variants more than classnames)

packages/components/chip/src/useChipElement.tsx
```diff
@@ -0,0 +1,83 @@
...
+          setIsPressed(!isPressed)
+          onClick && onClick(event, { pressed: isPressed as boolean })
+        },
+        ...(disabled && { disabled: true }),
+      },
+      closeIconProps,
+    ]
+  }
+
+  return [
```
acd02, line 76: Maybe it would be more legible to return an object instead of a tuple, i.e.:
```tsx
return { element: ChipElement, chipProps: elementProps, closeIconProps: closeIconProps }
```

packages/components/chip/src/Chip.styles.tsx
andresin87:
```suggestion
        no: ['max-w-sz-240'],
```

packages/components/chip/src/Chip.styles.tsx
```diff
@@ -0,0 +1,183 @@
...
+        mode: 'default',
+        hasCloseIcon: 'no',
+        isBordered: 'no',
+        class: tw(['px-md']),
+      },
+      {
+        mode: 'default',
+        hasCloseIcon: 'no',
+        isBordered: 'yes',
+        class: tw(['px-[7px]']),
```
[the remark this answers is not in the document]
Powerplex: Okay :) @andresz1 had a similar issue on the textarea PR, maybe for those cases we could use `calc()` to use the appropriate token minus a delta ? It will be more scalable that way.

packages/components/chip/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "vite build"
+  },
+  "dependencies": {
+    "@spark-ui/icon": "^1.7.4",
+    "@spark-ui/icons": "^1.14.2",
```
kikoruiz, line 16: Won't it cause a circular dependency?

packages/components/chip/src/Chip.tsx
```diff
@@ -0,0 +1,118 @@
...
+        })}
+        {...{
+          ...chipProps,
+          ...otherProps,
+        }}
+        data-spark-component="chip"
+      >
+        <span
+          className={chipContentStyles({
+            isBordered: design === 'dashed' ? 'yes' : 'no',
```
kikoruiz, line 89: Why not using booleans here? 🤔

packages/components/chip/src/useChipElement.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: {
+  onClick?: (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+    args: { pressed: boolean }
+  ) => void
+  onClose?: React.MouseEventHandler<HTMLSpanElement>
+  asChild?: boolean
+  pressed?: boolean
+  defaultPressed?: boolean
+  disabled: boolean | null
```
kikoruiz, line 21: `null`?

kikoruiz (approved):

Issue #539, opened 2023-04-03
[Components] Button loading state


andresz1:
### Description

Buttons should provide a loading state option to let the users know that something is happening

### Problem Statement/Justification

I believe it's important to include this feature because it provides users with immediate feedback, even if the desired result is not immediately visible. This way, users are aware that the system is processing their request and can continue to interact with it without feeling uncertain or confused.

### Proposed Solution or API

Two props:
- `isLoading` will display a spark spinner
- `loadingText` optionally set some text when the button is in the loading state

```jsx
<>
<Button isLoading>
  Email
</Button>

<Button isLoading loadingText="Submitting">
  Submit
</Button>
</>
```

![Screen Shot 2023-04-03 at 13 08 23](https://user-images.githubusercontent.com/6877967/229493626-d22a6c54-4a29-4e57-987b-3a967794b038.png)

andresin87:
loading text maintains the button with given with default (non-liading) children?

PR #880, opened 2023-06-08
539 components button loading state

Powerplex:
**TASK**: #539

### Types of changes
- [x ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations

https://github.com/adevinta/spark/assets/2033710/b55435e3-c2d2-421c-8e03-4b7d72bce473

### DIFF
--- a/packages/components/button/package.json
+++ b/packages/components/button/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@spark-ui/internal-utils": "^1.6.1",
     "@spark-ui/slot": "^1.5.2",
+    "@spark-ui/spinner": "^0.2.8",
     "class-variance-authority": "0.5.2"
   },
   "peerDependencies": {
--- a/packages/components/button/src/Button.stories.tsx
+++ b/packages/components/button/src/Button.stories.tsx
@@ -1,4 +1,3 @@
-import { StoryLabel } from '@docs/helpers/StoryLabel'
 import { Checkbox } from '@spark-ui/checkbox'
 import { Icon } from '@spark-ui/icon'
 import { Check } from '@spark-ui/icons/dist/icons/Check'
@@ -37,7 +36,7 @@ export const Sizes: StoryFn = _args => (
   <div className="gap-md flex flex-wrap items-center">
     {sizes.map(size => {
       return (
-        <Button key={size} size={size} className="!bg-error">
+        <Button key={size} size={size}>
           Button {size}
         </Button>
       )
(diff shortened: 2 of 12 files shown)


packages/components/spinner/src/Spinner.styles.tsx
```diff
@@ -11,6 +11,7 @@ export const spinnerStyles = cva(
   {
     variants: {
       size: {
+        current: ['w-[1em]', 'h-[1em]'],
```
andresz1, line 14: `u-current-font-size` instead ? This way we avoid adding extra hard coded clases 🙏🏻

packages/components/icon-button/src/IconButton.doc.mdx
```diff
@@ -59,9 +60,25 @@ Use `intent` prop to set the color intent of a button.
 
 <Canvas of={stories.Intent} />
 
+## Loading
+
+Use `isLoading` prop to set the color intent of a button.
```
andresz1, line 65:
```suggestion
Use the isLoading prop to show its loading state. By default, the button will show a spinner and leave the button's width unchanged
```

packages/components/button/src/Button.stories.tsx
```diff
@@ -103,10 +105,44 @@ export const Icons: StoryFn = _args => (
   </div>
 )
 
+export const Loading: StoryFn = () => {
+  const [isLoading, setIsLoading] = useState(true)
+
+  return (
+    <div className="gap-lg flex flex-col">
+      <Checkbox checked={isLoading} onClick={() => setIsLoading(!isLoading)}>
```
andresz1, line 113: This is a personal opinion but I would try to keep the stories as simple as possible and make then atomic. This way developers can easily copy/paste and our visual regression tests would be more granular

packages/components/button/src/Button.stories.tsx
```diff
@@ -35,7 +37,7 @@ export const Sizes: StoryFn = _args => (
   <div className="gap-md flex flex-wrap items-center">
     {sizes.map(size => {
       return (
-        <Button key={size} size={size}>
+        <Button key={size} size={size} className="!bg-error">
```
andresz1, line 40: Is the className needed here ?

packages/components/button/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@spark-ui/internal-utils": "^1.6.1",
     "@spark-ui/slot": "^1.5.2",
+    "@spark-ui/spinner": "^0.2.8",
```
acd02, line 17: Unrelated to your PR, but it seems odd that the spinner package is still at version ~0.

Is it not considered stable yet?

kikoruiz (approved):

packages/components/slot/src/Slot.tsx
```diff
@@ -8,3 +8,19 @@ export type SlotProps = PropsWithChildren<React.HTMLAttributes<HTMLElement>>
 export const Slot = forwardRef<HTMLElement, SlotProps>((props, ref) => {
   return <SlotPrimitive ref={ref} {...props} />
 })
+
+/**
+ * When using Radix `Slot` component, it will consider its first child to merge its props with.
+ * In some cases, you might need to wrap the top child with additional markup without breaking this behaviour.
+ */
+export const wrapPolymorphicSlot = (
+  asChild: boolean | undefined,
```
acd02, line 17:
```suggestion
  asChild?: boolean,
```

packages/components/button/src/Button.tsx
```diff
@@ -10,14 +12,32 @@ export interface ButtonProps
...
+  isLoading?: boolean
+  /**
+   * If your loading state should only display a spinner, it's better to specify a label for it (a11y).
+   */
+  loadingLabel?: string
+  /**
+   * If your loading state should also display a label, you can use this prop instead of `loadingLabel`.
+   * **Please note that using this can result in layout shifting when the Button goes from loading state to normal state.**
+   */
+  loadingText?: string
```
acd02, line 27: Does it make sense to pass both props (loadingLabel and loadingText) simultaneously?

If not, shouldn't we ensure (using TS) that it's not possible to do so?

packages/components/slot/src/Slot.tsx
Powerplex: Semantically it is not the same. Inside of a function it means the first parameter is optional but the next one it not. I want it to be mandatory but allow `undefined` (which is different)
Capture d’écran 2023-06-08 à 16 31 16

packages/components/button/src/Button.tsx
Powerplex: You should pass one or the other.

It's a nightmare to do this kind of thing with TS. It means a huge union type of two interfaces.

acd02: It's not that bad actually 😛,
```tsx
type CustomProps =  | {
      asChild?: boolean
      isLoading?: boolean
      loadingText?: string
      loadingLabel?: never
    }
  | {
      asChild?: boolean
      isLoading?: boolean
      loadingLabel?: string
      loadingText?: never
    }

type ButtonProps = PropsWithChildren<
  Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'disabled'>
> &
  ButtonStylesProps &
  Propse
```

acd02: But I understand if you think it's not worth adding this TS complexity, as it doesn't add that much value in the end 👍

packages/components/slot/src/Slot.tsx
acd02: 💅, when a function takes more than two arguments, it may be more legible to use an object as the argument instead

acd02 (approved):

andresin87 (approved):

Issue #765, opened 2023-05-13
Base Textarea Component


andresin87:
### Description

The `TextArea` component should render the native `Html TextArea` only.
The package will be named `@spark-ui/textarea`

take a look on #760 

### Problem Statement/Justification

We need to spread and decouple all the pieces in a very small parts to be able to compose all of them.

### Proposed Solution or API

The way it will be uses is for self using that element for composing custom renders:

### ***Scenario 7***: TextBlock look like a multiline TextField as a TextBlockField
```jsx
<TextBlockField>textarea label</TextBlockField>
```
this will be equivalent to:
```jsx
<Input.Group>
	<FloatingLabel>
		<TextArea /> <--this is what we need
	</FloatingLabel>
</Input.Group>
```

### Additional Information

https://mui.com/base/react-textarea-autosize/

andresz1:
https://github.com/adevinta/spark/pull/881

soykje:
As PR has been merged, couldn't we close this issue now? cc @andresz1

andresz1:
Yup

PR #881, opened 2023-06-08
Feat textarea

andresz1:
## Textarea

**TASK**: #765 

This PR:
- Add initial textarea implementation

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshot

Screen Shot 2023-06-09 at 11 47 58


packages/components/input/src/InputElement.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { cva, VariantProps } from 'class-variance-authority'
 
 export const inputElementStyles = cva(
-  ['text-neutral pointer-events-none absolute top-1/2 -translate-y-1/2'],
+  ['text-neutral pointer-events-none absolute top-[22px] -translate-y-1/2'],
```
andresz1, line 4: I need to improve this part. Fixed values shouldn't be used

Powerplex: If it can't be eliminated, maybe at least we could use `calc()` to use the appropriate spacing token minus 2px to reach the same result. It would scale better the day we change token values :)

packages/components/textarea/src/Textarea.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+      <Icon>
+        <Check />
+      </Icon>
+    </InputGroup.RightElement>
+  </InputGroup>
+)
+
+export const FieldLabel: StoryFn = _args => {
+  return (
+    <FormField className="!gap-sm" name="title">
```
Powerplex, line 86: Just a detail but I think we should use default `gap` of FormField in our stories OR challenge the default gap value if it's better with a small one by default :)

packages/components/textarea/src/Textarea.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export interface TextareaProps
+  extends ComponentPropsWithoutRef<'textarea'>,
+    Pick<InputProps, 'intent'> {}
+
+export const Textarea = forwardRef<HTMLTextAreaElement, PropsWithChildren<TextareaProps>>(
+  ({ className, intent, disabled, rows = 1, ...others }, ref) => {
+    return (
+      <Input
+        className={cx(className, 'py-md resize-y')}
```
Powerplex, line 13: I'm wondering if we should not add the `self-start` class here (`align-self: flex-start`) because the component renders differently when used as a standalone VS inside of a FormField, where it takes the full width available:
Capture d’écran 2023-06-14 à 18 11 10

WDYT ?

Powerplex (approved):

packages/components/textarea/src/Textarea.stories.tsx
andresz1: I agree. I will try to discuss this with the design team

andresz1: I will discuss this today 🙌🏻

andresz1: Discussed with the design team I will create another issue to perform the change

packages/components/input/src/InputElement.styles.ts
andresz1: Yup. Gonna ping @acd02 to check if he has a another idea

kikoruiz (approved):

packages/components/input/src/InputElement.styles.ts
acd02: I don't have a better idea for handling that use case 😄, so lgtm

PR #883, opened 2023-06-09
feat(kbd): Keyboard Key

andresin87:
**TASK**: 

A component for add shortkey placeholder to Inputs fields. Also available to document A11y story sections

### Types of changes

### Screenshots - Animations

https://chakra-ui.com/docs/components/kbd/usage


Powerplex (commented):
Not a big fan of the package name "kbd". It's not explicit what it is :)

Maybe `KeyboardKey` or `Keyboard` ?

packages/components/kbd/src/Kbd.styles.tsx
```diff
@@ -0,0 +1,5 @@
+import { cva } from 'class-variance-authority'
+
+export const kbdStyles = cva([
+  'border-sm border-b-md bg-neutral-container ps-md pe-md border-outline whitespace-nowrap rounded-sm font-mono text-[0.8em] font-bold leading-none',
```
Powerplex, line 4: "text-[0.8em]" is not a typography token of spark. 

I think we should stick to spark tokens. Also text is not vertically centered (maybe because no line-Height is set as it is an hardcoded value and not a token ?)
Capture d’écran 2023-06-09 à 10 14 27

packages/components/kbd/src/Kbd.tsx
```diff
@@ -0,0 +1,9 @@
+import { ComponentPropsWithoutRef, forwardRef, PropsWithChildren } from 'react'
+
+import { kbdStyles } from './Kbd.styles'
```
Powerplex, line 3: Imho it's unnecessary to split styles from markup for simpler components like this one. This is one of the benefit of using Tailwind.

packages/components/kbd/src/Kbd.test.tsx
```diff
@@ -0,0 +1,27 @@
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { describe, expect, it, vi } from 'vitest'
+
+import { Kbd } from './Kbd'
+
+describe('Kbd', () => {
```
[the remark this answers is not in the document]
andresin87: dont know what to test cuz its a visual only element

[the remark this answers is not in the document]
Powerplex: Remove the tests then :) Only keep the first one but replace "Hello world !" by "Ctrl" or something like that :)

packages/components/kbd/src/Kbd.styles.tsx
andresin87: `leading-3`?

Powerplex: I don't think we have a font-size named like that. Maybe `caption` or `small` ?

packages/components/kbd/src/Kbd.tsx
andresin87: I prefer being very strict with the DS rules and decoupling the view from the logic. WDYT?

andresin87: done!

kikoruiz (approved):

Issue #893, opened 2023-06-12
Update FormField state management


andresz1:
### Description

FormField should provide an API to set success and alert statuses

### Problem Statement/Justification

Success and alert status for input can not be set easily

### Proposed Solution or API

```
<FormField status="error">
  <Input />
  <FormField.StatusMessage status="error">Error</FormField.StatusMessage>
  <FormField.StatusMessage status="alert">Alert</FormField.StatusMessage>
  <FormField.StatusMessage status="success">Success</FormField.StatusMessage>
</FormField>
```

andresz1:
https://github.com/adevinta/spark/pull/916

PR #916, opened 2023-06-16
Feat validation state

andresz1:
## Improve form field validation state

**TASK**: #893 

This PR improve form field validation state by replacing `isInvalid` by `state`.

The `state` optional prop could be:
- `error`
- `success`
- `alert`

The `FormField.ErrorMessage` is replace by `FormField.StateMessage` that displays conditionally a message based on the `state` prop 

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

Screen Shot 2023-06-16 at 15 19 18


andresin87 (approved):

packages/components/form-field/src/FormFieldStateMessage.tsx
```diff
@@ -0,0 +1,34 @@
...
+export interface FormFieldStateMessageProps
+  extends FormFieldMessageProps,
+    FormFieldStateMessageStylesProps {}
+
+export const FormFieldStateMessage = forwardRef<HTMLSpanElement, FormFieldStateMessageProps>(
+  ({ className, state, ...others }, ref) => {
+    const field = useFormField()
+
+    if (field.state !== state) {
+      return null
```
soykje, line 19: If I understand correctly the point here would be to avoid incoherent states? But why having to declare a state on this component, if it relies on FormField context? :thinking:

soykje: I was thinking about a third option, I guess:
```tsx
<FormField status="error">
  <Input />
  <FormField.StatusMessage status={status.token}>{status.message}</FormField.StatusMessage>
</FormField>
```

soykje: Well I was thinking more about the second option, but maybe we also could consider another one:
```tsx
const defaultStatusMessages = {
 error: 'Something went wrong',
  success: 'Great! Message sent!',
  alert: 'Warning, youre house is under fire...',
}

<FormField status="error" statusMessages={defaultStatusMessages}>
  <Input />
  <FormField.StatusMessage />
</FormField>
```
...where the message itself would be extracted from `defaultStatusMessages` (this prop could have default value for sake of simplicity). Wdyt?

andresz1: Agree 👍🏻 I think that we can keep both APIs but in the examples only illustrate the second one. I added the three components perhaps tests and example could be added in another PR

soykje (approved):

Issue #902, opened 2023-06-13
[Components] - Popover design review and updates


Powerplex:
Waiting for Popover specs to proceed to design review and/or updates. 

In the meantime component is published as a beta version and located in the experimental section.

thomas-leguellec:
To check with @turolopezsanabria and @mathieuhervouetlbc

PR #935, opened 2023-06-21
Popover close button

Powerplex:
**TASK**: #902 

- Adding `Popover.Close` subcomponent
- Adjust padding of `Popover.Content` when `Popover.Close` is used
- Set max width of the Popover

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations

Popover with close button:
Capture d’écran 2023-06-21 à 20 46 48

Popover with close button (keyboard navigation):
Capture d’écran 2023-06-21 à 20 46 57


packages/components/popover/src/PopoverClose.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Icon } from '@spark-ui/icon'
+import { IconButton } from '@spark-ui/icon-button'
+import { Close as CloseSVG } from '@spark-ui/icons/dist/icons/Close'
+import { cx } from 'class-variance-authority'
+import { forwardRef, useEffect } from 'react'
+
+import { usePopover } from './PopoverContext'
+
+export type CloseProps = RadixPopover.PopoverCloseProps & {
+  'aria-label': string
```
soykje, line 11: Why do we need to add this to type? Isn't it already defined by default?

packages/components/popover/src/PopoverClose.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const Close = forwardRef<HTMLButtonElement, CloseProps>(
+  ({ 'aria-label': ariaLabel, className, ...rest }, ref) => {
+    const { setHasCloseButton } = usePopover()
+
+    useEffect(() => {
+      setHasCloseButton(true)
+
+      return () => setHasCloseButton(false)
+    })
```
soykje, line 22:
```suggestion
    }, [])
```
Wdyt?

packages/components/popover/src/PopoverClose.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { cx } from 'class-variance-authority'
+import { forwardRef, useEffect } from 'react'
+
+import { usePopover } from './PopoverContext'
+
+export type CloseProps = RadixPopover.PopoverCloseProps & {
+  'aria-label': string
+}
+
+export const Close = forwardRef<HTMLButtonElement, CloseProps>(
```
andresz1, line 14: IMO this component should be named `CloseIconButton`

packages/components/popover/src/PopoverClose.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+      <RadixPopover.Close
+        data-spark-component="Popover.Close"
+        ref={ref}
+        className={cx('absolute top-md right-md', className)}
+        asChild
+        {...rest}
+      >
+        <IconButton size="sm" intent="neutral" design="ghost" aria-label={ariaLabel}>
```
andresz1, line 32: - I think that `aria-label` could be part of `rest`
- Why not making `size`, `intent` and `design` configurable ? and with those values by default ?

packages/components/popover/src/PopoverClose.tsx
```diff
@@ -0,0 +1,42 @@
...
+export type CloseProps = RadixPopover.PopoverCloseProps & {
+  'aria-label': string
+}
+
+export const Close = forwardRef<HTMLButtonElement, CloseProps>(
+  ({ 'aria-label': ariaLabel, className, ...rest }, ref) => {
+    const { setHasCloseButton } = usePopover()
+
+    useEffect(() => {
+      setHasCloseButton(true)
```
andresz1, line 19: Isn't this going to cause a layout shift ?

packages/components/popover/src/Popover.doc.mdx
```diff
@@ -43,6 +44,8 @@ A basic popover should have a trigger element and a content element.
 The trigger is the element that should toggle the popover.
 The content is the popover itself.
 
+**Note: For headings inside Popover, please use the `header` HTML tag instead of `h1, h2`, etc.**
```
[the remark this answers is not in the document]
Powerplex: I discussed having a `Popover.Header` and `Popover.Footer` with Mathieu to explain to him why I think it's unnecessary (if we add one, it should be visible on the "Anatomy" section of the ZeroHeight documentation too)

If it is part of the anatomy specs should reflect it :) 

Basically it would look like this:
```
<Popover>
        <Popover.Trigger asChild>
          <Button>Trigger popover</Button>
        </Popover.Trigger>
        <Popover.Portal>
          <Popover.Content>
            <Popover.Header>Title</Popover.Header>
            <p>Are you sure you want to have that cookie now ?</p>
            <Popover.Arrow />
            <Popover.Close aria-label="Close the popover" />
          </Popover.Content>
        </Popover.Portal>
      </Popover>
```

Then should `Popover.Content` be a flex-col with a gap between elements then ? Or should the `Popover.Header` have a margin below by default ?

packages/components/popover/src/Popover.test.tsx
```diff
@@ -44,20 +44,20 @@ describe('Popover', () => {
       )
 
       // ...by default, only trigger elements are rendered
-      expect(document.querySelector('[data-spark-component="popover-anchor"]')).toBeInTheDocument()
-      expect(document.querySelector('[data-spark-component="popover-trigger"]')).toBeInTheDocument()
+      expect(document.querySelector('[data-spark-component="Popover.Anchor"]')).toBeInTheDocument()
```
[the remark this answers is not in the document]
Powerplex: I opened a thread to discuss this point on Slack using this PR :) I worked on the Chrome extension rencently and think it was weird to show "popover-anchor" instead of the displayName of the component inside the extension. I'll edit back to kebab-case.

packages/components/popover/src/PopoverClose.tsx
Powerplex: I need to make it mandatory for a11y

packages/components/popover/src/PopoverClose.tsx
Powerplex: I wanted to stick with Radix naming.

packages/components/popover/src/PopoverClose.tsx
Powerplex: It is not part of the specs. And making the size configurable would be extremely tricky as the padding-right of `Popover.Content` must be adjusted when the `Popover.Close` is rendered to avoid overlapping.

packages/components/popover/src/PopoverClose.tsx
soykje: Didn't we agree yesterday to avoid having that attribute mandatory?

packages/components/popover/src/Popover.doc.mdx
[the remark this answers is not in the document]
soykje: Ok then we should also update the docs :+1:

packages/components/popover/src/PopoverClose.tsx
Powerplex: Ok then I'll update the doc too

soykje (approved):

packages/components/popover/src/PopoverClose.tsx
andresz1: I think that is better to name it like that some developers know that is an `IconButton`

packages/components/popover/src/PopoverClose.tsx
andresz1: I think that with `useLayoutEffect` the problem is going to still happen but we can review together later on.

andresz1 (approved):

packages/components/popover/src/PopoverClose.tsx
Powerplex: Why do they need to know it is an `IconButton` that we chose to use ? I think it may be too descriptive.

`Popover.CloseButton` maybe ? Its true that `Close` as a name might be too short and can be mistaken for an action/verb.

- Most design have no close button for Popovers
- The one who does either name it `CloseButton` (ex: Chakra) or `Close` (Radix)

PR #937, opened 2023-06-22
feat(close-button): add new element

andresin87:
**TASK**: 

### Types of changes

### Screenshots - Animations

Why `<div role="button"/>`?:
it might be placed anywhere (including over buttons) and it warns that you should not place a button over a button

Why key handler?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#sect1


packages/components/close-button/src/CloseButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+      },
+      [onClick]
+    )
+
+    const onKeyUpHandler = useCallback((event: React.KeyboardEvent<HTMLDivElement>) => {
+      ;['Enter'].includes(event.key) && buttonRef.current?.click()
+    }, [])
+
+    return (
+      <span
```
Powerplex, line 35: I am confused about the wrappers of this component: `span > div`.

Why do we need two level ? Why is the div with a role `button` not a `button` (html) ?

In the popover I derivated the close button from an IconButton directly, maybe we can do that ?

packages/components/close-button/src/CloseButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+          cursor: disabled ? 'disabled' : 'pointer',
+          className,
+        })}
+        onClick={onCloseHandler}
+        {...props}
+      >
+        <div
+          ref={buttonRef}
+          role="button"
+          {...(disabled ? { 'aria-disabled': disabled } : { tabIndex: 0 })}
```
Powerplex, line 47: If it were a normal `button` the `disabled` attribute already manages this behaviour.

packages/components/close-button/src/CloseButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+          onKeyUp={onKeyUpHandler}
+          className={closeButtonIconStyles({ cursor: disabled ? 'disabled' : 'pointer' })}
+        >
+          <Icon label={label} className="opacity-dim-3">
```
Powerplex, line 51: I have seen with designers and it should be neutral token not opacity on the icon :) They updated the figma for it on the popover for example. Maybe it needs to be updated for the Chip too ?

packages/components/close-button/src/CloseButton.tsx
andresin87: yes, its not for adding it or not, its because it is only required to be placed only if its disabled, so i dnt want to add more html properties than the required ones

packages/components/close-button/src/CloseButton.styles.tsx
```diff
@@ -0,0 +1,34 @@
...
+    defaultVariants: {
+      cursor: 'pointer',
+    },
+  }
+)
+
+export const closeButtonIconStyles = cva(
+  [
+    'inline-flex rounded-full',
+    'focus-visible:ring-outline-high ring-inset focus-visible:outline-none focus-visible:ring-2',
```
Powerplex, line 21: I think for the Popover the styles are different. I am not sure we should have a dedicated package for a CloseButton at this stage, maybe it could be part of `Input` or `TextField` to be re-used by TextAreaField and later on we could deep dive into this component ?

I think for Drawer/Dialog it will be yet another design.

andresin87: the chip and the input had the same one and will be very common

https://mantine.dev/core/close-button/

packages/components/close-button/src/CloseButton.tsx
andresin87: there is not an iconbutton:

the clicking area for the closing action is bigger than the icon per-se, the componnet is preventing the action to go up cuz it can be placed inside of another button (chip) thats why its a div roled

Issue #886, opened 2023-06-09
Add readOnly styles to input


andresz1:

andresin87:
Only on the input? there are no readOnly styles for the radio, checkbox, switch elements?

andresz1:
For now the scope of this task is only for the input component

Powerplex:
We changed a lot of code for the Input package and will re-apply what you did @kikoruiz on the newest version :) I move it back to "in progress".

PR #986, opened 2023-07-03
feat(input): add readonly styles

kikoruiz:
## feat(input): add readonly styles

**TASK**: #886

Add read only styles to the current `Input` and `TextField` components.

### Types of changes

- [x] 💄 Styles

### Screenshots - Animations

read only


Powerplex:
@kikoruiz me and @andresz1 had a meeting with designer about TextField + Input.

We will update it in a way that TextField will be fused with Input to keep a single package and design/specs will be updated to keep things simple :) We will work on it tomorrow morning if you want to sync with us.

packages/components/text-field/src/TextFieldFloatingLabel.styles.ts
```diff
@@ -1,7 +1,14 @@
 import { cva, VariantProps } from 'class-variance-authority'
 
 export const textFieldFloatingLabelStyles = cva(
-  ['absolute', 'transition-all', 'duration-200', 'cursor-auto'],
+  [
+    'absolute',
+    'transition-all',
+    'duration-200',
+    'cursor-auto',
+    'peer-disabled:text-on-surface/dim-3',
```
kikoruiz, line 9: Missing style for disabled state.

kikoruiz:
Not valid anymore

Issue #929, opened 2023-06-19
Add state icons to InputGroup


andresz1:
### Description

Add state icon elements to InputGroup for:
- Error
- Alert
- Success

![Image](https://github.com/adevinta/spark/assets/6877967/14de3240-820c-4d1c-bce3-350c7799d071)

### Problem Statement/Justification

Provide feedback about the input state with icons

PR #997, opened 2023-07-04
Input group status

Powerplex:
**TASK**: #929 

- `TextField` has been removed. We will only have one `Input` package. The only feature that differentiated both package was the floating label, which we will implement here in the next PR.
- We clarified with designers that we will have "solid" addons in the Input. For multiple reasons:
    -  focus style (outline) should remain around the input itself not its addons (because they can be also focusable individually)
    - it makers the floating label easier to manage (won't overlap with addons)
    - Make it more easy to distinguish them from leading/trailing icons (non-interactive)
- State indicator (outline colors and icon for `error, alert` and `success` states) are managed in the component to enforce a11y (colorblind people needs at least an icon in addition to the outline color) 

### Types of changes
- [x] 🧾 Documentation
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
Capture d’écran 2023-07-04 à 15 57 56


packages/components/input/src/Input.stories.tsx
```diff
@@ -37,60 +40,100 @@ export const Disabled: StoryFn = _args => (
...
   return (
     <InputGroup>
-      <InputGroup.LeftAddon>https://</InputGroup.LeftAddon>
-      <Input aria-label="Website" />
-      <InputGroup.RightAddon>.com</InputGroup.RightAddon>
+      <Input type={isVisible ? 'text' : 'password'} />
+
+      <InputGroup.TrailingAddon className="px-md">
+        <IconButton
+          className="!text-body-1"
```
acd02, line 93: Why use a `!` modifier here? Shouldn't you set a size of md instead?

acd02 (approved):

andresz1 (approved):

PR #1001, opened 2023-07-06
Feat input addon

Powerplex:
**TASK**:  #1000 

This PR is to refine the styles of addons, especially focus and hover styles, and their combination with disabled state.

We make use of `ring-inset` and replaced border by `ring` (box-shadow) for a better look and feel.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations

https://github.com/adevinta/spark/assets/2033710/d8931b33-f1df-46ed-8826-37f75a252450

### DIFF
--- a/packages/components/input/src/InputLeadingAddon.tsx
+++ b/packages/components/input/src/InputLeadingAddon.tsx
@@ -7,7 +7,11 @@ export type InputLeadingAddonProps = InputAddonProps
 
 export const InputLeadingAddon = forwardRef<HTMLDivElement, InputLeadingAddonProps>(
   ({ className, ...others }, ref) => (
-    <InputAddon ref={ref} className={cx(className, 'rounded-l-lg border-r-none')} {...others} />
+    <InputAddon
+      ref={ref}
+      className={cx(className, 'rounded-l-lg !rounded-r-none mr-[-1px]')}
+      {...others}
+    />
   )
 )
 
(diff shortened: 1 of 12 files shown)


andresz1 (approved):

packages/components/input/src/InputLeadingAddon.tsx
```diff
@@ -7,7 +7,11 @@ export type InputLeadingAddonProps = InputAddonProps
 
 export const InputLeadingAddon = forwardRef<HTMLDivElement, InputLeadingAddonProps>(
   ({ className, ...others }, ref) => (
-    <InputAddon ref={ref} className={cx(className, 'rounded-l-lg border-r-none')} {...others} />
+    <InputAddon
+      ref={ref}
+      className={cx(className, 'rounded-l-lg !rounded-r-none mr-[-1px]')}
```
acd02, line 12: I noticed you're using a negative margin of 1px here. Have you tested to make sure it's reliable and won't break when zooming in or out, stuff like that?

Powerplex: This negative margin is to make the right shadow of the input overlap with the left shadow of the addon. That way when focus goes from the input to the input it looks seamless. 

It works well when zooming in too.

acd02 (approved):

Issue #887, opened 2023-06-09
[Components] Input - clear button


andresz1:
### Description

Input component should be able to provide an easy way to remove its content with a button

### Problem Statement/Justification

For example, it could be really useful when using it as search input

### Proposed Solution or API

TBD

Powerplex:
I started a branch for it. Pending because we think we could get the package out of experimental when addons are ready, and keep the clearButton feature for after we release the stable version.

PR #1004, opened 2023-07-07
feat(input): clear button

Powerplex:
**TASK**: #887

This PR add the clear button:

- It’s optional
- Not focusable with the tab key
- When is clicked besides removing the value of the input it focus it
- It’s displayed automatically when the value of the input is not empty
- Custom click handlers to the button could be added
- When the input is focused and escape key is pressed the value of the input is removed
- Accessible name could be set by consumers

https://github.com/adevinta/spark/assets/6877967/c6983195-6710-4bc1-bfca-4df89b8a788c

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles


turolopezsanabria (approved):

turolopezsanabria:
Maybe not for this PR, but the border-radius is not ok in Chrome for iPhone 

IMG_8286

andresz1:
> Maybe not for this PR, but the border-radius is not ok in Chrome for iPhone
> 
> IMG_8286

Thanks @turolopezsanabria I already opened a PR to fix it #1026

andresin87:
image
image
image

I think there is too much space

andresin87:
if the <Input /> gets a type="srearch" attr, what is the behavior given? Could we set it and hide the native clear icon?

andresz1:
> image image image
> I think there is too much space

It's because of the "safety area" but we still need to check the spacing with the design team

andresz1:
> if the gets a type="srearch" attr, what is the behavior given? Could we set it and hide the native clear icon when including it as a prop or detecting the ClearButton inside ?

Yes! you are right. We need to hide it. We can do it in another PR

andresz1 (approved):

andresin87:
I dnt know why we are using the <IconButton> with that space. I think we agree this should be fully clickable last wednesday. Can you please confirm if i'm wrong?

kikoruiz (approved):

turolopezsanabria:
Sorry for being late: Yes it's correct @andresin87, the area should be fully clickable

> I don't know why we are using that space. I think we agree this should be fully clickable last Wednesday. Can you please confirm if I'm wrong?

Issue #1005, opened 2023-07-07
[Chore] - setup nx cache


Powerplex:
Bumbed Lerna to v7, which allows us to use NX by simply optin to `"useNx": true,`.

For now I only setup nx cache for the lerna build. That way locally when we build it only builds updated packages :) Way faster than before:

![Image](https://github.com/adevinta/spark/assets/2033710/f8a2f580-6401-49b0-a24e-eebc18c81351)

acd02:
I think we can close this issue, right? cc @Powerplex

soykje:
There is no opened PR there, is it?

PR #1006, opened 2023-07-07
chore: setup nx cache with lerna for better DX

Powerplex:
**TASK**: #1005 

The goal is to improve our workflow by making the build faster (only locally updated package are built)

### Types of changes

- [x] 🛠️ Tool


package.json
```diff
@@ -99,8 +99,9 @@
     "eslint-plugin-testing-library": "5.10.3",
     "husky": "8.0.3",
     "jsdom": "21.1.1",
-    "lerna": "6.4.1",
+    "lerna": "^7.1.1",
```
andresin87, line 102: Top!

kikoruiz (approved):

Issue #996, opened 2023-07-04
[RadioGroup] Styles


soykje:
After first LBC implementations feedbacks we noticed some style issues on `RadioGroup` component:

- [ ] Input width should not shrink
- [ ] Vertical alignment between input and label (with label max fontSize recommendation) should be as defined in the specs
- [ ] Add multiline content examples

soykje:
The label max fontSize recommendation would be not to exceed 24px lineHeight. In other words, not using title sizes (which seems logical with component itself).

Issue #994, opened 2023-07-04
[Checkbox] Styles


soykje:
After first LBC implementations feedbacks we noticed some style issues on `Checkbox` component:

- [x] Input width should not shrink
- [x] Vertical alignment between input and label (with label max fontSize recommendation) should be as defined in the specs
- [x] Add multiline content examples

soykje:
The label max fontSize recommendation would be not to exceed 24px lineHeight. In other words, not using title sizes (which seems logical with component itself).

PR #1008, opened 2023-07-10
[FIX] multiline labeled inputs

kikoruiz:
## fix(checkbox|radio-group) multiline labeled inputs

**ISSUES**: #994 #996

Fix multiline labeled inputs as defined in specs.

### Types of changes

- [x] 💄 Styles

### DIFF
--- a/packages/components/radio-group/src/RadioGroup.doc.mdx
+++ b/packages/components/radio-group/src/RadioGroup.doc.mdx
@@ -39,6 +39,10 @@ import { RadioGroup } from '@spark-ui/radio-group'
 
 <Canvas of={RadioStories.Default} />
 
+## With a multiline label
+
+<Canvas of={RadioStories.WithAMultilineLabel} />
+
 ## Using a FormField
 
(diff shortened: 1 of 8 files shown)


packages/components/radio-group/src/RadioGroup.doc.mdx
```diff
@@ -39,6 +39,10 @@ import { RadioGroup } from '@spark-ui/radio-group'
 
 <Canvas of={RadioStories.Default} />
 
+## With a multiline label
+
```
soykje, line 43: Don't you think we should be more explicit about the limitations concerning label line-height? As said within original issue it could seem obvious and out of our scope to care about these bad implementations...
But it might also be cool to add a small reminder about it (not using headings sizes for labels). Wdyt?

acd02 (approved):

packages/components/radio-group/src/RadioGroup.doc.mdx
kikoruiz: Yep, adding it is worth it 😉

Powerplex (approved):

andresin87 (approved):

PR #1010, opened 2023-07-12
refactor(checkbox): improve docs on multiline labels for checkbox and radio

soykje:
**TASK**: #1009 

To help consumer having better implementations of our components, we want to improve our documentation about the multiline usecase.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
image


soykje:
Also added a small update on radio indicator styles, as the initial `66.6667%` was bringing some rounding issues (based on how the browser can do its maths...)

packages/components/checkbox/src/Checkbox.doc.mdx
```diff
@@ -28,11 +28,9 @@ import { Checkbox, CheckboxGroup } from '@spark-ui/checkbox'
 
 ## Usage
 
-<Canvas of={CheckboxStories.Default} />
-
-## With a multiline label
+Checkbox label can be multiline. Nevertheless you should take care of not using heading font sizes, to preserve alignments.
```
acd02, line 31: I think we could maybe explain explicitly why we should avoid using headings for labels, i.e., the element used for the label should not have a greater line height than the height of the checkbox. Because if the label ends up being in multiple lines, it will break the vertical alignment

soykje: I was not sure about that... I didn't want to put some too specific information there: checkbox/radio sizes could evolve in time. But this recommendation would still be relevant, wdyt?

acd02: Additionally, maybe it would be helpful to showcase an example where the label has a greater line-height than the checkbox, so that the negative effects of this misuse can be observed 🤷

> But this recommendation would still be relevant, wdyt?

I don't know because what are "heading font sizes"? I am not sure consumers will understand what we are referring to

soykje: Hmmm... you might be right, I'll try to be more explicit :+1:

soykje:
```
Checkbox label can be multiline. Still you should care of not using elements (for the label) with greater line height than the checkbox input height, to preserve vertical alignment.
```
Wdyt?

acd02: LGTM, and wdyt about adding an example that illustrates what it would look like if a label with too much line height were used incorrectly (a misuse example)?

acd02 (approved):

packages/components/checkbox/src/Checkbox.doc.mdx
soykje: I'm afraid this could lead us to too much complexity on the documentation... :thinking: This could completely be discussed, obviously, but IMHO we should keep it simple, and have snippets only for API examples, as for now. Wdyt?

acd02: yeah, alright. We can maybe revisit this idea later if we sense there's a need for it

Issue #1039, opened 2023-07-18
[Components] - Radio/Checkbox - making label full width


Powerplex:
Update the radio and checkbox components to allow for the label to take up the full width of the container

PR #1048, opened 2023-07-19
feat: update radio/checkbox styles to allow full label width

acd02:
**TASK**: #1039 

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [x] 💄 Styles


acd02:
@Powerplex , we initially thought that we should only apply the `flex-grow: 1` style when the orientation is set to **vertical**. 
However, it turns out that it doesn't have any negative impact on the components, even when the orientation is horizontal. Therefore, there's no need to add a condition to our code

Powerplex (approved):

kikoruiz (approved):

Issue #1040, opened 2023-07-18
[Components] - Radio/Checkbox - adding stories for custom implementation


Powerplex:
- radio/checkbox wrapped by `@spark-ui/label`
- radio/checkbox wrapped by `@spark-ui/label` + input hidden using `@spark/visually-hidden`

PR #1049, opened 2023-07-19
docs(checkbox/radio group): add implementation and guidance for custom checkbox/radio group components

acd02:
**TASK**: #1040

Add an implementation of a custom checkbox component to the documentation that can provide guidance to developers who need to implement a custom checkbox / radio group

### Types of changes
- [ ] 🛠️ Tool
- [x] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/checkbox/src/Checkbox.stories.tsx
+++ b/packages/components/checkbox/src/Checkbox.stories.tsx
@@ -1,10 +1,13 @@
 import { FormField } from '@spark-ui/form-field'
 import { Close } from '@spark-ui/icons/dist/icons/Close'
 import { Plus } from '@spark-ui/icons/dist/icons/Plus'
+import { Label } from '@spark-ui/label'
+import { VisuallyHidden } from '@spark-ui/visually-hidden'
 import { Meta, StoryFn } from '@storybook/react'
+import { cx } from 'class-variance-authority'
 import { useState } from 'react'
 
-import { Checkbox } from './Checkbox'
+import { Checkbox, CheckboxProps } from './Checkbox'
 import { CheckboxGroup } from './CheckboxGroup'
 
 const meta: Meta<typeof Checkbox> = {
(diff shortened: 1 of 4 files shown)


packages/components/checkbox/src/Checkbox.stories.tsx
```diff
@@ -173,3 +176,99 @@ export const GroupImproved: StoryFn = () => {
...
+    const { value } = others
+
+    return (
+      <Label
+        id={id}
+        htmlFor={value}
+        className={cx(
+          'flex flex-wrap shadow rounded-md gap-md p-lg max-w-sz-320',
+          checked ? 'bg-success/dim-4' : '',
+          'focus-within:ring-1 focus-within:ring-outline-high',
```
Powerplex, line 192: I would remove the `ring-1` here. Guideline for spark focus is 
- `focus-visible` (keyboard nav only) set a ring-2.
- `focus-within` with a hover `ring-1` is only for inputs where the user can type.


Either remove it or replace it with `focus-visible:ring-2`

Powerplex (approved):

packages/components/radio-group/src/RadioGroup.stories.tsx
```diff
@@ -116,6 +119,118 @@ export const Disabled: StoryFn = _args => (
...
+    const id = useId()
+    const { value } = others
+
+    return (
+      <Label
+        id={id}
+        htmlFor={value}
+        className={cx(
+          'flex flex-wrap shadow rounded-md gap-md p-lg max-w-sz-320',
+          value === selectedValue ? 'bg-success/dim-4' : '',
```
kikoruiz, line 191: Not sure if we should write some agreement about this kind of logic when defining classnames 🤔

kikoruiz (approved):

packages/components/radio-group/src/RadioGroup.stories.tsx
acd02: What kind of agreements do you have in mind?

Issue #930, opened 2023-06-19
Create TextareaField component


andresz1:
### Description

Create TextareaField component using TextField component

### Problem Statement/Justification

There is no Textarea component with a floating label

Powerplex:
Adding the blocked label as long as Input is not approved

turolopezsanabria:
hello
Do you have a PR I can check?

andresz1:
Oops! I missed this comment. Yes I do have. https://github.com/adevinta/spark/pull/1053

PR #1053, opened 2023-07-21
Feat/textarea check

andresz1:
## Textarea update

**TASK**: #930 

This PR:
- Update `Textarea` component introducing `TextareaGroup`
- Update `InputGroup` component to use `id` instead of `displayName` to render children. This decouples the need of using input display names
- Update documentation

Notes:
- Once this PR is merged I think `Textarea` could be ready to be used
- Test related with form field in the text area does not work 😞 we are checking why

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles


packages/components/input/src/InputClearButton.tsx
```diff
@@ -47,4 +47,8 @@ export const InputClearButton = forwardRef<HTMLButtonElement, InputClearButtonPr
   }
 )
 
+export const InputClearButton = Object.assign(Root, {
+  id: 'ClearButton',
+})
```
Powerplex, line 52: I'm still not familiar with assigning an `id` to the component like this instead of a displayName. Can you summarize it please ? :)

Powerplex (approved):

kikoruiz (approved):

packages/components/input/src/InputClearButton.tsx
andresz1: Yes. I did this change to not be coupled with the `displayName`. For example, to add the `Textarea` I would have needed to go to the `InputGroup` and add support for it
```tsx
const input = findElement('Input', 'Textarea')
```
Doing it this way no changes need to be performed in the `InputGroup` and consumer can also created their own component that can work nicely with the `InputGroup` is needed extending the `id`.

I was inspired by chakra [InputGroup](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/input/src/input-group.tsx#L44) to perform this change

PR #1072, opened 2023-07-25
feat(input): input readonly styles

kikoruiz:
## feat(input): input readonly styles

**TASK**: #886

Add readonly specific styles to `input` component.

### Types of changes

- [x] 💄 Styles

### Screenshots - Animations
Screenshot 2023-07-27 at 15 10 19


turolopezsanabria:
I see this proposal very well.

Can we change the cursor of the read-only to `default`?
I know the read-only obviously needs to be focusable, but the cursor `text` gives me the impression that the input is "writable"

kikoruiz:
> I see this proposal very well.
> 
> Can we change the cursor of the read-only to `default`? I know the read-only obviously needs to be focusable, but the cursor `text` gives me the impression that the input is "writable"

@turolopezsanabria We can switch to the default cursor value, we can check, I was just suggesting what other DSs are doing like for example: https://mui.com/material-ui/react-text-field/#form-props

packages/components/input/src/Input.stories.tsx
```diff
@@ -123,6 +124,103 @@ export const Disabled: StoryFn = _args => {
...
+          defaultValue="Hello world"
+        />
+      </div>
+
+      <div>
+        <StoryLabel>Addons - solid</StoryLabel>
+
+        <InputGroup className="max-w-sz-320" readOnly={isReadOnly}>
+          <InputGroup.LeadingAddon asChild>
+            <IconButton intent="primary" design="filled" aria-label="Search">
```
Powerplex, line 152: It just needs to be adapted to the new colors :)

Powerplex (approved):

Issue #1082, opened 2023-07-26
[Components] - Dialog - Radix compound mapping and basics


Powerplex:
To initiate this component in experimental stage, I suggest we expose a compound component mapped over what Radix offers.

Then we will only have to add styles over it, as well as missing subComponent such as `Header/Body/Footer`

PR #1083, opened 2023-07-26
1030 components dialog experimental

Powerplex:
**TASK**: #1082 

Setup the compound for the `Dialog` component, using Radix.

I created an Epic for the `Dialog` to follow progress.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test

### DIFF
--- a/packages/components/dialog/src/Dialog.tsx
+++ b/packages/components/dialog/src/Dialog.tsx
@@ -1,7 +1,36 @@
-import { ComponentPropsWithoutRef, forwardRef, PropsWithChildren } from 'react'
+import type { FC } from 'react'
 
-export type DialogProps = ComponentPropsWithoutRef<'div'>
+import { CloseButton } from './DialogCloseButton'
+import { Content } from './DialogContent'
+import { Description } from './DialogDescription' // aria-describedby
+import { Overlay } from './DialogOverlay'
+import { Portal } from './DialogPortal'
+import { Root, type RootProps } from './DialogRoot'
+import { Title } from './DialogTitle' // aria-labelledby
+import { Trigger } from './DialogTrigger'
 
-export const Dialog = forwardRef<HTMLDivElement, PropsWithChildren<DialogProps>>((props, ref) => {
-  return <div ref={ref} {...props} />
+Trigger.displayName = 'Dialog.Trigger'
+Portal.displayName = 'Dialog.Portal'
+Overlay.displayName = 'Dialog.Overlay'
+Content.displayName = 'Dialog.Content'
+CloseButton.displayName = 'Dialog.CloseButton'
+Title.displayName = 'Dialog.Title'
+Description.displayName = 'Dialog.Description'
+
+export const Dialog: FC<RootProps> & {
+  Trigger: typeof Trigger
+  Portal: typeof Portal
+  Overlay: typeof Overlay
+  Content: typeof Content
+  CloseButton: typeof CloseButton
+  Title: typeof Title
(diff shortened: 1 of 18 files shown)


andresz1 (approved):

packages/components/dialog/src/Dialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Root, type RootProps } from './DialogRoot'
+import { Title } from './DialogTitle' // aria-labelledby
+import { Trigger } from './DialogTrigger'
+
+Trigger.displayName = 'Dialog.Trigger'
+Portal.displayName = 'Dialog.Portal'
+Overlay.displayName = 'Dialog.Overlay'
+Content.displayName = 'Dialog.Content'
+CloseButton.displayName = 'Dialog.CloseButton'
+Title.displayName = 'Dialog.Title'
```
soykje, line 17: Wouldn't we wrap the title and description from content within a `Heading`component, for ex.? Same for content itself...?

packages/components/dialog/src/DialogCloseButton.tsx
```diff
@@ -0,0 +1,12 @@
+import * as RadixDialog from '@radix-ui/react-dialog'
+import { type ReactElement } from 'react'
+
+export type CloseButtonProps = RadixDialog.DialogCloseProps
+
+export const CloseButton = ({ children, ...rest }: CloseButtonProps): ReactElement => (
+  <RadixDialog.Close data-spark-component="dialog-close-button" {...rest}>
```
soykje, line 7: Do we really want `data-spark-component` on each subcomponent from a compound one? Just asking, because it wasn't the case previously (see `Tabs`)

packages/components/dialog/src/index.ts
```diff
@@ -0,0 +1 @@
+export { Dialog } from './Dialog'
```
soykje, line 1: Don't we want to also export interfaces and types?

packages/components/dialog/src/Dialog.tsx
Powerplex: It is planned I just created the ticket for this :) https://github.com/adevinta/spark/issues/1101

Basically `Popover.Header` will contain the `Popover.Title` but can also contain other elements. So I think we need the 2 of them.

You can also wrap a title with `VisuallyHidden` when you don't need a Header: https://www.radix-ui.com/docs/primitives/components/dialog#title

So I think we need both :) But Radix does not offer Header/Body/Footer that we need, So I created a ticket.

packages/components/dialog/src/DialogCloseButton.tsx
Powerplex: I'm not sure, it depends on what we want to do with our future Chrome extension. I think we at least need to be able to visually detect the Trigger VS the Popover.Content. I can remove the reste maybe.

packages/components/dialog/src/Dialog.tsx
soykje: Oh great! Even if I'm a bit afraid now, of the amount of subcomponents we would end with :stuck_out_tongue_closed_eyes:

packages/components/dialog/src/DialogCloseButton.tsx
soykje: There is no problem there, but I guess we should all have an agreement on this, to keep our strategy consistent :+1:

soykje (approved):

PR #1084, opened 2023-07-26
🐤 Chip-stage-6

andresin87:
**TASK**: 

- Documentation
	- Definition extended
	- Filter demos changed to emojis
	- Improved how disabled state affects clearbutton
- ClearButton
	- Tabindex
	- Label

### Types of changes

- [x] 🧾 Documentation


packages/components/chip/src/Chip.doc.mdx
```diff
@@ -11,14 +12,23 @@ import * as stories from './Chip.stories'
...
+<Callout kind="warning">
+  Multiple chips should appear together in a group, whereas there should be no more than 3 buttons
+  in a single arrangement.
+</Callout>
+
+Chips are reactive and contextual, whereas buttons are static and predetermined.
+A chip should offer a different action depending on the nature of the content it supports,
+whereas a button should be a persistent fixture of a layout.
+From a user-flow perspective,
+chips represent forking paths in an experience while a button represents a linear step along a flow.
```
Powerplex, line 31: This block of text is too complicated imho. Terms like "forking paths" or "persistent fixture", etc are confusing. We should either remove this from the documentation (guidelines are in ZH) or keep it short and simple.

packages/components/chip/src/Chip.doc.mdx
```diff
@@ -11,14 +12,23 @@ import * as stories from './Chip.stories'
...
 Help people enter information, make selections, filter content, or trigger actions.
 
-Use chips to show options for a specific context.
+### Chips aren’t buttons
 
-Four types of usages:
+<Callout kind="warning">
+  Chips and buttons are similar in that they both provide visual cues to prompt users to take
+  actions and make selections. However, there are important distinctions to be aware of.
+</Callout>
```
Powerplex, line 20: Missing spacing between the 2 callouts

packages/components/chip/src/Chip.stories.tsx
```diff
@@ -152,13 +172,37 @@ export const UnionFilter: StoryFn = _args => {
...
+      '🦭': ['sea', 'land', 'wild'],
+      '🦆': ['sea', 'land', 'air', 'wild'],
+      '🐡': ['sea', 'wild'],
+      '🐟': ['sea', 'wild'],
+      '🐞': ['land', 'air', 'wild'],
+      '🐢': ['land', 'sea', 'wild', 'domestic'],
+      '🦜': ['land', 'air', 'wild', 'domestic'],
+    })
+  )
+).sort(() => 0.5 - Math.random())
```
Powerplex, line 198: We should keep in mind people will read the code ("show source") so we should keep it simple. Can you remove the random ? I think it will also be an issue for chromatic.

Powerplex: I think we don't need to use a combination of `Array.from`, `Math.random`, `new Map` and `Object.entries`, we could just have this: 
```
const animals = [
  { emoji: '🦍', tags: ['land', 'wild']},
  { emoji: '🦒', tags: ['land', 'wild']},
  { emoji: '🦌', tags: ['land', 'wild']},
  { emoji: '🐄', tags: ['land', 'domestic']},
  { emoji: '🐕', tags: ['land', 'domestic']},
  { emoji: '🐬', tags: ['sea', 'wild']},
  { emoji: '🐓', tags: ['land', 'domestic']},
  { emoji: '🐿️', tags: ['land', 'wild']},
  { emoji: '🦧', tags: ['land', 'wild']},
  { emoji: '🐳', tags: ['sea', 'wild']},
  { emoji: '🦈', tags: ['sea', 'wild']},
  { emoji: '🦭', tags: ['sea', 'land', 'wild']},
  { emoji: '🦆', tags: ['sea', 'land', 'air', 'wild']},
  { emoji: '🐡', tags: ['sea', 'wild']},
  { emoji: '🐟', tags: ['sea', 'wild']},
  { emoji: '🐞', tags: ['land', 'air', 'wild']},
  { emoji: '🐢', tags: ['land', 'sea', 'wild', 'domestic']},
  { emoji: '🦜', tags: ['land', 'air', 'wild', 'domestic']},
]

```

Maybe we can also only keep 2 examples (single selection and multiple selection)

Powerplex (changes requested):

packages/components/chip/src/Chip.stories.tsx
```diff
@@ -32,13 +36,29 @@ export default meta
...
+
+export const SingleSelectionFilter: StoryFn = () => {
+  const [activeFilter, setActive] = useState<undefined | string>('fruit')
 
   return (
     <div className="flex flex-col gap-md">
       <div className="flex flex-row items-start gap-md">
-        {['vowel', 'consonant', 'number', 'symbol'].map(filter => {
+        {['fruit', 'tubercle'].map(filter => {
```
acd02, line 61: it seems weird to me to handle filtering within the `map` method. Imho, it would be more appropriate to first filter the list and then map over it

packages/components/chip/src/Chip.stories.tsx
```diff
@@ -32,13 +36,29 @@ export default meta
...
+    Object.entries({
+      '🥝': ['fruit'],
+      '🍎': ['fruit'],
+      '🍇': ['fruit'],
+      '🍌': ['fruit'],
+      '🍆': ['tubercle'],
+      '🧅': ['tubercle'],
+      '🥔': ['tubercle'],
+      '🌶': ['tubercle'],
+      '🥕': ['tubercle'],
```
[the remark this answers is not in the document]
andresin87: I think fruits are vegetables too. Is it? I dont know

packages/components/chip/src/Chip.stories.tsx
andresin87: well probably the error is about the naming. Maybe better replace filter variable name to element, cuz the filter is the usage of the elements as children of the Chip.

packages/components/chip/src/Chip.doc.mdx
```diff
@@ -103,3 +145,31 @@ They enable user input and verify that input by converting text into chips.
...
+
+Suggestion chips help narrow a user’s intent by presenting dynamically generated suggestions, such as possible responses or search filters.
+
+<Canvas of={stories.Suggestion} />
+
+Suggestion chips can offer quick-reply options in a chat or email app, or It can help the user start a search.
+
+## Max Length
+
+The max length of any Chip is 240px including its borders, icons, and clear button.
```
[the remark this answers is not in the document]
andresin87: because it might be Max With, sorry 👌. I took it from the definition but yes. It's better to adapt the language for dev usages

PR #1098, opened 2023-07-27
Feat/resize textarea

andresz1:
## Disable textarea resize

**TASK**: #1099

This PR:
- Add `isResizable` prop to textarea
- Missing `data-spark-component` attributes

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles


packages/components/textarea/src/Textarea.tsx
```diff
@@ -2,12 +2,26 @@ import { Input } from '@spark-ui/input'
...
 const Root = forwardRef<HTMLTextAreaElement, PropsWithChildren<TextareaProps>>(
-  ({ className, disabled, rows = 1, ...others }, ref) => {
+  ({ className, disabled, rows = 1, isResizeDisabled = false, ...others }, ref) => {
     return (
-      <Input className={cx(className, 'py-[var(--sz-10)] resize-y')} disabled={disabled} asChild>
+      <Input
+        className={cx(
+          className,
+          'py-[var(--sz-10)]',
+          isResizeDisabled ? 'resize-none' : 'resize-y'
```
andresz1, line 19: Not `100%` sure about this one.
- Is the name of prop ok ?
- Should we let them control this by using `className` or is it better a `prop` like this ?

soykje: :wave: I would vote for leaving the possibility to developper to toggle resize on textarea through custom styles (`className`), but I guess this also could be discussed with design team? Elseway I'd prefer maybe a simpler prop name such as `isResizable` wdyt?

andresz1: I like `isResizable` better. However the downside is that the default value should be `true`. That's the reason why I chose `isResizeDisabled`. Perhaps we can discuss this during the day

soykje: Hmmm got it, I didn't understood that we wanted to keep the resize feat. by default :sweat_smile: Well, maybe we should discuss first with design team, and then see together :+1:

Powerplex (approved):

packages/components/textarea/src/Textarea.tsx
acd02: I vote for `isResizable` as well 👍

andresin87: and if we call it

`resizable`: boolean or vertical/horizontal: default true

andresin87 (approved):

soykje (approved):

turolopezsanabria (approved):
👏🏼

PR #1109, opened 2023-07-28
fix(form-field): fix undefined values when accessing length property

kikoruiz:
`FormField.CharactersCount`: The component throws and error when `value` prop is `undefined`.

### Types of changes

### DIFF
--- a/packages/components/form-field/src/FormField.test.tsx
+++ b/packages/components/form-field/src/FormField.test.tsx
@@ -246,4 +246,39 @@ describe('FormField error message', () => {
     expect(inputEl).toHaveValue('Hello')
     expect(charsCountEl).toBeVisible()
   })
+
+  it('should render the characters count component properly when `value` is undefined', () => {
+    const MAX_LENGTH = 6
+    const DEFAULT_VALUE = undefined
+
+    render(
+      <FormField name="text">
+        <FormField.Label>Text</FormField.Label>
+
+        <FormField.Control>
+          {({ id, name }) => (
+            <input
+              type="text"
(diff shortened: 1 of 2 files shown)


andresz1 (approved):

andresz1:
Should we add a test for this case ?

kikoruiz:
> Should we add a test for this case ?

Yes indeed ;)

packages/components/form-field/src/FormField.test.tsx
```diff
@@ -246,4 +246,39 @@ describe('FormField error message', () => {
...
+          <FormField.CharactersCount value={DEFAULT_VALUE} maxLength={MAX_LENGTH} />
+        </div>
+      </FormField>
+    )
+
+    const inputEl = screen.getByLabelText('Text')
+    const charsCountEl = screen.getByText('0/6')
+
+    expect(inputEl).toHaveValue('')
+    expect(charsCountEl).toBeVisible()
```
Powerplex, line 282: Personal opinion but I think it is easier to read assertions without variables

```
expect(screen.getByLabelText('Text')).toHaveValue('')
expect(screen.getByText('0/6')).toBeVisible()
```

Powerplex (approved):

packages/components/form-field/src/FormField.test.tsx
kikoruiz: Agree, I was based on a different example where it makes more sense, not here ;)

andresin87 (commented):
✅

PR #1133, opened 2023-07-31
1132 components alertdialog

andresz1:
## AlertDialog

**TASK**: #1132 

This PR add a first iteration of the `AlertDialog` component

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

https://github.com/adevinta/spark/assets/6877967/af08fcc7-beea-4085-b6ce-e19010f63ee2


packages/components/alert-dialog/package.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "vite build"
+  },
+  "dependencies": {
+    "class-variance-authority": "0.4.0",
+    "@spark-ui/dialog": "^1.2.1",
```
Powerplex, line 16: Oops I forgot to set `dialog` to `0.*` when I created it :(

packages/components/alert-dialog/src/AlertDialogAction.tsx
```diff
@@ -0,0 +1,13 @@
+import { Dialog, DialogCloseProps } from '@spark-ui/dialog'
+import { ElementRef, forwardRef } from 'react'
+
+export type AlertDialogActionElement = ElementRef<typeof Dialog.Close>
+export type AlertDialogActionProps = DialogCloseProps
+
+export const AlertDialogAction = forwardRef<AlertDialogActionElement, AlertDialogActionProps>(
+  (props, ref) => {
+    return <Dialog.Close ref={ref} data-spark-component="alert-dialog-action" {...props} />
```
Powerplex, line 9: I think this is a mistake to use `Dialog.Close` for an `Action` subcomponent ? 

Aside from that I think we don't need to expose `AlertDialog.Action` because it does not add anything to the button. I checked on Radix, it doesn't add any html attribute or behaviour.

packages/components/alert-dialog/src/AlertDialog.doc.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+## Import
+
+```tsx
+import { AlertDialog } from "@spark-ui/alert-dialog"
+```
+
+## Props
+
+<ArgTypes of={AlertDialog} />
```
Powerplex, line 28: Do you want to add the `description` and `subComponents` here now ?

packages/components/alert-dialog/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+export { type AlertDialogBodyProps } from './AlertDialogBody'
+export { type AlertDialogCancelProps } from './AlertDialogCancel'
+export { type AlertDialogContentProps } from './AlertDialogContent'
+export { type AlertDialogDescriptionProps } from './AlertDialogDescription'
+export { type AlertDialogFooterProps } from './AlertDialogFooter'
+export { type AlertDialogHeaderProps } from './AlertDialogHeader'
+export { type AlertDialogOverlayProps } from './AlertDialogOverlay'
+export { type AlertDialogPortalProps } from './AlertDialogPortal'
+export { type AlertDialogTitleProps } from './AlertDialogTitle'
+export { type AlertDialogTriggerProps } from './AlertDialogTrigger'
```
Powerplex, line 28: You might not need this whole block of type imports, as you already export them here: `packages/components/dialog/src/index.ts`

packages/components/alert-dialog/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const AlertDialog: FC<AlertDialogProps> & {
```
Powerplex, line 30: Don't you also need to set the `displayNames` before exporting the compound, like we do in the Dialog ?
```
Trigger.displayName = 'Dialog.Trigger'
Portal.displayName = 'Dialog.Portal'
Overlay.displayName = 'Dialog.Overlay'
Content.displayName = 'Dialog.Content'
CloseButton.displayName = 'Dialog.CloseButton'
Title.displayName = 'Dialog.Title'
Description.displayName = 'Dialog.Description'
```

packages/components/alert-dialog/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "access": "public"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "vite build"
+  },
+  "dependencies": {
+    "class-variance-authority": "0.4.0",
```
andresin87, line 15:
```suggestion
    "class-variance-authority": "0.5.2",
```

Powerplex: We should also update CVA in dialog and tabs then :)

andresz1: Done

packages/components/alert-dialog/src/index.ts
andresz1: IMO one thing is the alert dialog props and another the dialog props and shouldn't be mixed

packages/components/alert-dialog/src/index.ts
andresz1: Yes :( this is needed only for storybook D: I think that we should avoid defining the `displayName` two times

andresin87 (approved):

packages/components/alert-dialog/src/index.ts
Powerplex: Oh I thought it was exported from the same package, ok

packages/components/alert-dialog/src/index.ts
Powerplex: I agree but it is important the the component's names are correct when a user click on "show source"

Powerplex (approved):

PR #1143, opened 2023-08-01
Proposal - Storybook table of content

Powerplex:
**TASK**: #1142 

See attached task.

### Types of changes
- [x] 🧾 Documentation

### Screenshots - Animations
Capture d’écran 2023-08-01 à 19 57 52


andresz1 (approved):

documentation/helpers/ToC/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+  useEffect(() => {
+    const list = document.querySelectorAll('h2, h3')
+
+    if ([...list].length) {
+      setHeadings(document.querySelectorAll('h2, h3'))
+    }
+  }, [])
+
+  console.log(headings)
```
acd02, line 32:
```suggestion

```

documentation/helpers/ToC/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const activeAnchor = useActiveAnchor(headings)
+
+  return headings ? (
```
acd02, line 35:
```suggestion
  if (!headings.length) return
  return headings.map(...)
```
or
```suggestion
  if (!headings) return
  return headings.map(...)
```

Instead of using an expanded ternary expression (which requires scanning through multiple lines of code to get to the other side of the ternary), maybe we could use an early return instead

documentation/helpers/ToC/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+    tagName: {
+      H2: 'pl-md text-body-1',
+      H3: "pl-[calc(theme('spacing.xl')*1)] text-body-2",
+      H4: "pl-[calc(theme('spacing.xl')*2)] text-caption",
+    },
+  },
+})
+
+export const ToC = () => {
+  const [headings, setHeadings] = useState(null)
```
acd02, line 22: This should be typed as `HTMLHeadingElement[]` imho

documentation/helpers/ToC/useActiveAnchor.js
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react'
```
acd02, line 1: Why not use TS? 😢

documentation/helpers/ToC/utils.js
```diff
@@ -0,0 +1,11 @@
+export function scrollToAnchor(anchorId) {
+  const element = document.getElementById(anchorId)
+  if (element) {
+    element.scrollIntoView({ behavior: 'smooth' })
+  }
+}
+
+export function updateUrlAnchor(hash) {
+  // eslint-disable-next-line no-undef
+  parent.location.href = parent.location.href.split('#')[0] + '#' + hash
```
acd02, line 10: Since this function performs side effects, it may be better to avoid this function and instead have the side effects directly at the call site where they are not hidden. This could avoid any potential confusion 🤷

documentation/helpers/ToC/useActiveAnchor.js
Powerplex: Because it was supposed to be a task that took 5 minutes by using SB's Table of Content feature. But unfortunately I couldn't, I'll update to TS :D

kikoruiz (approved):

PR #1231, opened 2023-08-08
fix: use performance observer to ensure to scroll only when main thread is not blocked

acd02:
**TASK**: #1183

use PerformanceObserver to ensure that scrollIntoView() is only called when the main thread is not blocked

### Types of changes
- [x] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### 📽️ Screens:

**before:**
before-one

**after:**
after-one


documentation/helpers/ToC/usePerformanceObserver.ts
```diff
@@ -0,0 +1,39 @@
...
+  useEffect(() => {
+    const observer = new PerformanceObserver(list => {
+      for (const entry of list.getEntries()) {
+        if (entry.entryType === 'longtask' && list.getEntries().length === 1) {
+          setTimeout(() => callbackRef.current(), 0)
+        }
+
+        if (entry.name === CUSTOM_MARK) {
+          performanceRef.current.clearMarks(CUSTOM_MARK)
+          setTimeout(() => callbackRef.current(), 500)
```
Powerplex, line 26: I understand the first `setTimeout(..., 0)` but not this one. 

Is it the call to `performanceRef.current.mark(CUSTOM_MARK)` that triggers the `PerformanceObserver` after the initial render to force a "re-sync" after 500ms in case the first one did not work ?

acd02: It's nothing but a big mess 🥲😢😅😔🙁😭

Inside the `if (entry.name === CUSTOM_MARK` block, the **callback prop** will be executed right away when the main thread is not blocked. However:

- I use `setTimeout` because, under the hood, Storybook calls `scrollIntoView` for the heading matching the hashtag present in the URL (if present)
- If the `callback prop` is executed at the same time as the `scrollIntoView` invocation from Storybook, we will run into problems because we cannot have two invocations of scrollIntoView at the same time.
- So, I use this `setTimeout` as a safety net to make sure the callback prop is called after the eventual invocation from Storybook.

And if the main thread is blocked, then the `if (entry.name === CUSTOM_MARK` will be executed, but nothing will happen (since the main thread is blocked). But the **callback prop** will be called once more when there is no blocking task left, inside the `entry.entryType === 'longtask' && list.getEntries().length === 1` block

andresin87 (approved):

PR #1233, opened 2023-08-08
🐤 Chip Stage 7

andresin87:
**TASK**: 

- Fix design with bug
- Apply stage 6 feedback
- Sort stories applying guidelines
- Add keyboard interaction definitions
- Change onClear API to get aligned with what is made in the input package

### Types of changes

### DIFF
--- a/packages/components/chip/package.json
+++ b/packages/components/chip/package.json
@@ -12,7 +12,7 @@
     "build": "vite build"
   },
   "peerDependencies": {
-    "@spark-ui/theme-utils": "latest",
+    "@spark-ui/theme-utils": "^3.0.2",
     "react": "^16.8 || ^17.0 || ^18.0",
     "react-dom": "^16.8 || ^17.0 || ^18.0",
     "tailwindcss": "^3.0.0"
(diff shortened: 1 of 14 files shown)


packages/components/chip/package.json
```diff
@@ -28,8 +28,10 @@
...
   "license": "MIT",
   "dependencies": {
-    "@spark-ui/icon": "latest",
-    "@spark-ui/icons": "latest",
-    "@spark-ui/slot": "latest"
+    "@spark-ui/icon": "^2.0.2",
+    "@spark-ui/icons": "^1.19.5",
+    "@spark-ui/internal-utils": "^2.0.2",
+    "@spark-ui/slot": "^1.6.1",
+    "emulate-tab": "^1.2.1"
```
Powerplex, line 35: I just ping @kikoruiz :) I am a bit lost with when to use `latest` or not. Has something changed regarding this ?

packages/components/chip/src/Chip.doc.mdx
```diff
@@ -50,44 +65,81 @@ import { Chip } from '@spark-ui/chip'
...
+<Canvas of={stories.Disabled} />
 
-## Assist
+### Max Width
+
+The max width of any Chip is 240px including its borders, icons, and clear button.
+
+<Canvas of={stories.MaxWidth} />
+
+## Accessibility
```
Powerplex, line 109: a11y section should be moved to the end of the file :)

packages/components/chip/package.json
andresin87: I realise `latest` only works in lerna CI publish step, but locally it does not infer to the current changed package (linking ints changes) and downloads the latest published

turolopezsanabria (commented):
👍🏼

Issue #991, opened 2023-07-03
[Component] Progress-bar


soykje:
Waiting for Design specs, we want to open the discussions about possible API for the Progress component ([here](https://github.com/adevinta/spark/discussions/992)) then after agreement start development in parallel with design team.

andresz1:
- Preview: https://sparkui.vercel.app/?path=/docs/experimental-progress--docs
- Specs: https://www.figma.com/file/0QchRdipAVuvVoDfTjLrgQ/Spark-Specs-and-Handoff?type=design&node-id=7229-1207&mode=design&t=PxwE4VjVC6bfL6MB-0

PR #1302, opened 2023-08-22
feat(progress): add label to progress component

andresz1:
## Progress

**TASK**: #991 

This PR:
- Add progress label
- Add progress indeterminate state
- Add progress documentation
- Add progress tests
- Fixes wrong dependency in `Popover` component

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Screenshots - Animations

https://github.com/adevinta/spark/assets/6877967/fe83894f-434e-4e4b-b750-6e045f67b030


packages/components/progress/src/Progress.tsx
```diff
@@ -1,29 +1,57 @@
...
+      value: valueProp,
+      max = 100,
+      isIndeterminate = false,
+      children = <ProgressBar />,
+      ...others
+    },
+    ref
+  ) => {
+    const id = useId(idProp)
+    const [labelId, setLabelId] = useState<string | undefined>()
```
acd02, line 33:
```suggestion
    const [labelId, setLabelId] = useState<string>()
```

packages/components/progress/src/ProgressContext.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react'
+
+export interface ProgressContextValue {
+  value: number
+  max: number
+  isIndeterminate: boolean
+  onLabelId: (id: string | undefined) => void
```
acd02, line 7:
```suggestion
  onLabelId: (id?: string) => void
```

packages/utils/tailwind-plugins/animations/index.js
```diff
@@ -137,12 +137,22 @@ module.exports = plugin.withOptions(
...
+            },
+          },
         },
         animation: {
           'slide-top': 'slideTop 0.5s ease-in-out',
           'slide-right': 'slideRight 0.5s ease-in-out',
           'slide-bottom': 'slideBottom 0.5s ease-in-out',
           'slide-left': 'slideLeft 0.5s ease-in-out',
+          'indeterminate-bar':
+            'indeterminateBar 1.5s cubic-bezier(0.65, 0.815, 0.735, 0.395) infinite normal none running',
```
andresin87, line 155: 👀 ⚠️

andresin87: for me better have it style inline than create a token that affects the whole DS and on the other package.

**S** - O - L - I - D

packages/components/progress/src/ProgressLabel.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export type ProgressLabelProps = ComponentPropsWithoutRef<'span'>
+
+export const ProgressLabel = forwardRef<HTMLSpanElement, ProgressLabelProps>(
+  ({ id: idProp, children, ...others }, forwardedRef) => {
+    const id = useId(idProp)
+    const { onLabelId } = useProgress()
+    const rootRef = useCallback(
+      (el: HTMLSpanElement) => {
+        onLabelId(el ? id : undefined)
```
acd02, line 15: I'm not sure I understand what's going on here, we have some `React.Context`that exposes a `onLabelId` function,
and you're calling that function here and passing it the generated id from the line above. But I'm not seeing where we're actually setting the id on the element itself 🤔

packages/utils/tailwind-plugins/animations/index.js
andresz1: ?

acd02: I agree with you @andresin87 , but I think we're kind of stuck here, imho.
Since we can't define a keyframe animation inline, and if we define it in its own CSS file, the consumers of the `Progress` component will have to download that CSS file as well 🫤

acd02: But maybe for cases like this, we could use a prefix, similar to the one used for utility classes such as `u-current-font-size` and `u-shadow-border-transition`

For example, something like `isolated-` or `standalone-` 🤷 
This would allow me to filter these out inside our TCV

packages/components/progress/src/ProgressLabel.tsx
acd02: Yes, I remember what you said yesterday about SSR and how React Spectrum was dealing with it, but I overlooked the `onLabelId: setLabelId` part 👍

packages/components/progress/src/Progress.tsx
```diff
@@ -1,29 +1,57 @@
...
+      ...others
+    },
+    ref
+  ) => {
+    const id = useId(idProp)
+    const [labelId, setLabelId] = useState<string>()
+
+    const value = useMemo(() => {
+      return { value: valueProp ?? 0, max, isIndeterminate, onLabelId: setLabelId }
+    }, [max, valueProp, isIndeterminate, setLabelId])
```
acd02, line 37: I thought I already left a comment here, but apparently not 😅  
Do we really need to use `useMemo`  here? Doesn't seem like a heavy computation

andresz1: I normally add it to avoid possible re-renders (see [post](https://blog.agney.dev/useMemo-inside-context/)) but I guess that in this won't be that important

packages/utils/tailwind-plugins/animations/index.js
andresz1: Which name would you prefer? I'm fine with either of them.

packages/components/progress/src/Progress.tsx
acd02: Yes, I try to stay away from `useMemo`, `useCallback`, `useEffect` unless I really need them, as it tends to create noise imho, but it's not a blocker for me

packages/utils/tailwind-plugins/animations/index.js
```diff
@@ -137,12 +137,22 @@ module.exports = plugin.withOptions(
             '0%': { transform: 'translateX(0)' },
             '100%': { transform: 'translateX(-100px)' },
           },
+          indeterminateBar: {
```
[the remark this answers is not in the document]
andresz1: Sure! will do

soykje (approved):

Issue #1256, opened 2023-08-10
[Components] - Button - loading button loses focus


Powerplex:
When clicking or pressing Enter on a button, if it goes into "loading" mode, it loses the focus because we set it as `disabled`.

We have seen other design system use:
- `aria-disabled` instead of the `disabled` HTML attribute
- `aria-busy="true"` 

We should figure which is best and implement it.

For now, the focus goes back to the top of the page, which breaks the user flow when navigating with the keyboard.

![image](https://github.com/adevinta/spark/assets/66770550/984f8ca3-1267-4760-ac43-05de9daaaa48)

soykje:
Seems clearer to use `aria-disabled` here. Also we may be interested in having a shared/common solution for all Spark components that could be disabled :+1:

PR #1304, opened 2023-08-22
refactor(button): improv. loading state focus management

soykje:
**TASK**: #1256 

Button loading state was previously linked to a disabled state. This caused focus loss that we don't want here. So we will update implementation using `aria-busy` attribute instead.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
image

### DIFF
--- a/packages/components/button/src/Button.tsx
+++ b/packages/components/button/src/Button.tsx
@@ -1,7 +1,7 @@
 import { Slot, wrapPolymorphicSlot } from '@spark-ui/slot'
-import { Spinner } from '@spark-ui/spinner'
+import { Spinner, type SpinnerProps } from '@spark-ui/spinner'
 import { cx } from 'class-variance-authority'
-import React, { MouseEvent, PropsWithChildren } from 'react'
+import React, { type MouseEvent, type PropsWithChildren } from 'react'
 
 import { buttonStyles, type ButtonStylesProps } from './Button.styles'
 
@@ -49,24 +49,28 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     ref,
   ) => {
     const Component = asChild ? Slot : 'button'
-    const isDisabled = !!disabled || isLoading
+
+    const shouldNotInteract = !!disabled || isLoading
 
     /**
      * When using `asChild` (polymorphic) it's possible that the button becomes another HTML element.
      * Depending on its tag, it could break the `disabled` html attribute preventing the clicks on a disabled button.
      */
     const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
-      if (isDisabled) event.preventDefault()
-      if (onClick) onClick(event)
+      if (shouldNotInteract) {
+        event.preventDefault()
+        // eslint-disable-next-line newline-before-return
+        return
+      }
+
+      onClick?.(event)
     }
 
-    const spinner = isLoading ? (
-      <Spinner
-        size="current"
-        className={loadingText ? 'inline-block' : 'absolute'}
-        {...(loadingLabel && { 'aria-label': loadingLabel })}
-      />
-    ) : null
+    const spinnerProps = {
+      size: 'current' as SpinnerProps['size'],
+      className: loadingText ? 'inline-block' : 'absolute',
+      ...(loadingLabel && { 'aria-label': loadingLabel }),
+    }
 
     return (
       <Component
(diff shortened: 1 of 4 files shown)


packages/components/button/src/Button.tsx
```diff
@@ -49,15 +49,21 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
     const Component = asChild ? Slot : 'button'
-    const isDisabled = !!disabled || isLoading
+
+    const shouldNotInteract = !!disabled || isLoading
 
     /**
      * When using `asChild` (polymorphic) it's possible that the button becomes another HTML element.
      * Depending on its tag, it could break the `disabled` html attribute preventing the clicks on a disabled button.
      */
     const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
```
acd02, line 59: Should we also handle other events, such as `onKeyDown`, `onKeyPress`etc...🙃🥲😢😭

soykje: Not sure to understand your remark... Could you precise your point plz?

acd02: Sure,
```tsx
<Button isLoading onKeyDown={() => fetch('someUrl')}>Default button</Button>
```

The `onKeyDown` handler will not be prevented

soykje: :crying_cat_face:

soykje: Code updated to take care of all (?) mouse and keyboar events that should be blocked while `isLoading`. Happy to discuss :+1:

packages/components/button/src/Button.tsx
```diff
@@ -48,49 +47,83 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
+        'mouseleave',
+        'mouseover',
+        'mouseout',
+        'keydown',
+        'keypress',
+        'keyup',
+        'submit',
+      ]
+
+      if (shouldNotInteract) {
```
acd02, line 86: I noticed that we're adding a lot of event listeners. I’m afraid this could maybe cause [performance issues](https://gomakethings.com/why-event-delegation-is-a-better-way-to-listen-for-events-in-vanilla-js/#web-performance), if there are many instances of the Button component on a page.

To avoid this, maybe we could instead create a lookup object that contains all the event handlers we need as keys, and then depending on the `shouldNotInteract` predicate, we could prevent the default event action, something like:
```tsx
// List of blocked event names that should not trigger interactions
const disabledEventsList = [
  'onClick',
  'onMouseDown',
  'onMouseUp',
  'onMouseEnter',
  'onMouseLeave',
  'onMouseOver',
  'onMouseOut',
  'onKeyDown',
  'onKeyPress',
  'onKeyUp',
  'onSubmit',
]

...

function makeDisabledEventHandlers() {
  return disabledEventsList.reduce((acc, eventName) => {
    // If no event handler was passed, skip
    if (!others[eventName]) return acc
		
    // If an event handler was passed, prevent its execution
    acc[eventName] = event => {
      event.preventDefault()
      event.stopPropagation()
    }

    return acc
  }, {})
}


...

<Component
  data-spark-component="button"
  disabled={!!disabled}
  {...others}
  {...(shouldNotInteract && makeDisabledEventHandlers())} // Override event handlers we are interested in, if those were passed
>
```

Bonus, it saves us from having to use `useEffect` and `useRef` 🧹

soykje: You're right. If I may another suggestion, wdyt about this:
```ts
const disabledEventHandlers = useMemo(() => {
  if (!shouldNotInteract) return

  return blockedEventHandlers.reduce((acc, eventHandler) => {
      acc[eventHandler] = undefined

      return acc
    },
    {} as Record<(typeof blockedEventHandlers)[number], (() => void) | undefined>,
  )
}, [shouldNotInteract])
```

acd02: Also, I think we can do without the `useMemo` , although I seem to be in the minority within the team, as everyone else seems to use it 😄

soykje: Ahah :smile: Well, it seemed fitting our needs to me, to help take care of `shouldNotInteract` variable without involving to many recalculations (as for ex. in a form we may toggle this state several times)

acd02 (approved):

Powerplex (approved):

PR #1345, opened 2023-08-28
⁉️ add trailinComma eslint config

andresin87:
⁉️ WDYT

**TASK**: 

During https://github.com/adevinta/spark/pull/1308 our prettier rules had changed from `es5` to `all`. Do you prefer the current default config or come back again to our lastest config but with the new prettier version?
https://prettier.io/docs/en/options.html#trailing-commas
### Types of changes

- [x] ⚙️  Config


andresin87:
Dnt wrry about the patch

acd02:
Since it is the new default since **Prettier v3**, I do not mind leaving this setting as is

However, if some people strongly oppose this and wish to revert back to es5, I will not oppose it 🤷

Powerplex (approved):

Powerplex:
I don't have a strong opinion about trailing/no trailing commas, I just like that the rule is set.

Issue #1374, opened 2023-09-01
[Components] Avoid animating overlay dialog when size is fullscreen


acd02:
When the size of the overlay dialog is set to fullscreen, the animation out effect should be avoided

acd02:
> [!NOTE]  
After discussing with @Powerplex and @soykje ,we have decided to put this PR on hold for now.
> We may discard this pull request and instead refactor the Dialog component to use React Context.
> This will allow us to handle the bug without resorting to the escape hatch used in the current PR (adding a class to the body)

elisa-hery:
Another PR was opened by @Powerplex, same goal but different approach --> https://github.com/adevinta/spark/pull/1410 
@Powerplex 's PR is tested & validated ✅

acd02:
Yes, closing in favor of #1410 👍

PR #1375, opened 2023-09-01
fix(dialog): avoid animating overlay dialog when size is fullscreen

acd02:
**TASK**: #1374

Avoid animating overlay dialog when size is fullscreen

> [!WARNING]  
After discussing with @Powerplex and @soykje ,we have decided to put this PR on hold for now.
> We may discard this pull request and instead refactor the Dialog component to use React Context.
> This will allow us to handle the bug without resorting to the escape hatch used in the current PR (adding a class to the body)

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/dialog/src/Dialog.test.tsx
+++ b/packages/components/dialog/src/Dialog.test.tsx
@@ -53,8 +53,8 @@ describe('Dialog', () => {
       </Dialog>
     )
 
-    const fullscreenClass = 'modal-is-fullscreen'
-    const animationPreventionSelector = '[body.modal-is-fullscreen_&]:[animation:none]'
+    const fullscreenClass = 'dialog-is-fullscreen'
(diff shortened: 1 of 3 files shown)


Powerplex (changes requested):
`modal` -> `dialog` (everywhere)

acd02:
After discussing with @Powerplex  and @soykje ,we have decided to put this pull request on hold for now.

We may discard this pull request and instead refactor the Dialog component to use React Context.
This will allow us to handle the bug without resorting to the escape hatch used in the current PR (adding a class to the body)

acd02:
handled by #1410

Issue #1395, opened 2023-09-05
Allow users to add custom Tailwind tokens


acd02:
Add a Tailwind plugin that enable users to add new custom tokens, such as an arbitrary height for a header for instance.

This will provide more flexibility and customization options for users.

PR #1408, opened 2023-09-07
feat(tailwind-plugins): add token extender plugin

acd02:
**TASK**: #1395

This PR adds a Tailwind plugin that will enable users to add new custom tokens, such as an arbitrary height for a header for instance.

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/documentation/using-spark/FAQ.mdx
+++ b/documentation/using-spark/FAQ.mdx
@@ -114,27 +114,29 @@ import { FAQ } from '@docs/helpers/FAQ'
       label="Can I extend the Spark theme with additional tokens?"
     />
       <FAQ.Answer>
-        If you require tokens that are not part of the Spark theme, you can use the [Tailwind API](https://tailwindcss.com/docs/theme#extending-the-default-theme) to achieve this.
-        
-        Since Spark uses Tailwind under the hood, you have access to all its APIs.
+        If you require tokens that are not part of the Spark theme, you can use our [token extender plugin](?path=/docs/utils-tailwind-plugins-token-extender--docs#tailwind-token-extender-plugin) to achieve this.
         
         For example, to add a new breakpoint, you can do the following:
         ```js
         // tailwind.config.js
 
+        const sparkPlugins = require('@spark-ui/tailwind-plugins')
+
         module.exports = {
-          content: [...],
-          plugins: [...],
-          theme: {
-            extend: {
-              screens: {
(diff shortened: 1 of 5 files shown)


Powerplex:
I think I prefer to keep the default Tailwind's implementation for this.

Once developers are familiar with Tailwind, it's easier if they know of the theme structure and set tokens themselves. I don't think we need this abstraction.

In leboncoin they have: 
```
module.exports = plugin.withOptions(
  () => undefined,
  () => ({
    theme: {
      extend: {
        screens: {
          tiny: '480px',
          custom: '971px',
        },
        width: {
          'page-min': '320px',
          'page-max': '1066px',
        },
        maxWidth: {
          'page-max': '1066px',
        },
        minWidth: {
          'page-min': '320px',
        },
      },
    },
  })
)
```

So `1066px` is used twice. They could simply use a constant to reach the same result:

```
const MIN_PAGE_WIDTH = '320px'
const MAX_PAGE_WIDTH = '1066px'

module.exports = plugin.withOptions(
  () => undefined,
  () => ({
    theme: {
      extend: {
        screens: {
          tiny: '480px',
          custom: '971px',
        },
        width: {
          'page-min': MIN_PAGE_WIDTH,
          'page-max': MAX_PAGE_WIDTH,
        },
        maxWidth: {
          'page-max': MAX_PAGE_WIDTH,
        },
        minWidth: {
          'page-min': MIN_PAGE_WIDTH,
        },
      },
    },
  })
)
```

acd02:
👋  @Powerplex ,

The idea for this plugin came from a question from @andresz1 , who wanted to use an arbitrary token for the new documentation website.

As you said, we can use the default method of extending the theme manually, which we already do. This plugin just makes things a little less cumbersome when you have a token that you want to use for different theme keys. However, beyond that, I agree that it does not add much value 🤷

The only other thing it does is create a CSS variable for you, which you can reference in your code using arbitrary values like `[flex-basis:--your-token]`.
But even then, you could achieve the same result without this plugin by using the theme function ➡️  `[flex-basis:theme(height.your-token)]`.

So, I would board with not using this plugin and closing the PR 👍

acd02:
Closing this PR, as we can get by without it for now

PR #1410, opened 2023-09-08
fix(dialog): do not render overlay in fullscreen size

Powerplex:

The fadeOut applied to the overlay of the dialog when it's closing was looking weird on `fullscreen` size.

I introduce a context so that the `Dialog.Overlay` is aware of the size passed to `Dialog.Content` and is never rendered in `fullscreen` size.

### Types of changes

### Screenshots - Animations

https://github.com/adevinta/spark/assets/2033710/27e12119-e7f8-4360-b3f6-82b7b6bcb786


packages/components/dialog/src/DialogContent.tsx
```diff
@@ -1,33 +1,44 @@
...
-        {...rest}
-      >
-        {children}
-      </RadixDialog.Content>
-    </div>
-  )
+  ): ReactElement => {
+    const { setIsFullScreen } = useDialog()
+
+    useLayoutEffect(() => {
```
acd02, line 20:
```suggestion
    useEffect(() => {
```

Powerplex: Why ? I prefer `useLayoutEffect` it's good for this case: https://react.dev/reference/react/useLayoutEffect

acd02: > Usage: Measuring layout before the browser repaints the screen
That is not the case here, you don't read values from an HTML element 🤷

Powerplex: No by I need this value updated in the state before the component is rendered, not after.

kikoruiz (approved):

packages/components/dialog/src/DialogContent.tsx
acd02: Yes, we can discuss it this morning 👍

acd02 (approved):

andresin87 (approved):

PR #1487, opened 2023-09-26
refactor(button): deprecate icon-button component

soykje:
To simplify the maintenance of our library we wanted to remove repetitive `IconButton` component, as it provides no real specificity in comparison to `Button` original one.

Basically the idea here would be to add a `hasIconOnly` prop that would enable specific styles.

### Types of changes
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
image


acd02 (commented):
I prefer `iconOnly` rather than `hasIconOnly`, as it feels more like HTML properties such as `read-only` (imho)

acd02 (approved):

soykje:
> I prefer `iconOnly` rather than `hasIconOnly`, as it feels more like HTML properties such as `read-only` (imho)

I don't have any strong opinion about that... I guess we can "vote" and decide which one we choose? :slightly_smiling_face:

Powerplex:
Can you update it to use semantic commit (name of the PR) + declare it as a proposal (draft ?) so that we discuss it (not all devs were in the discussion to deprecate `IconButton`)

PR #1488, opened 2023-09-26
fix(checkbox): set parent element's position to relative

warpdesign:
## fix(checkbox): set parent element's position to relative

**TASK**: #1489

Since this was not set element could be appear in any parent whose position not static and so could mess up with the layout.

Setting the main div's position to `relative` makes sure it doesn't get rendered in a random div (ie: the first parent whose position is set).


Powerplex (approved):

andresin87:
thks for @warpdesign

PR #1502, opened 2023-09-27
feat(text-link): enabled disabling of underline on text link

Powerplex:

Because the component is using TW `underline`  class, we must use `!` ("important") on every class.. so for the user to have a TextLink without underline, but underline on hover/focus, they must do:

```
className="!no-underline hover:!underline focus:!underline"
```

I think this is very error-prone because we take the risk that they do this:
```
className="!no-underline"
```

So I edit to have a cleaner way to disable it: 
```
<TextLink underline={false} ...
```

### Types of changes
- [x] 💄 Styles

### Screenshots - Animations

https://github.com/adevinta/spark/assets/2033710/957af60a-58e3-4698-a07f-2436613db3b6


turolopezsanabria (approved):
thnx!

andresin87 (approved):

PR #1508, opened 2023-09-28
feat(button): PROPOSAL - toggle button example

Powerplex:

I worked on multiple ways to approach toggle buttons, noticed a few things:

1. We will not have specific design specs for every intent/design for toggle buttons (`Button` and `IconButton`). 
2. Because we don't have any design, we can't make use of `defaultPressed` to apply any style that would make it obvious to the user that a button is pressed.
3. We are left with only `pressed` and `onPressedChange`. Pressed is just an alias for `aria-pressed`. `onPressedChange` is simply called on `onClick` internally to toggle the boolean.
4. `Toggle` from radix is not meant to be used as an HoC, it's a proper/distinct component. We are not meant to customize it into an HoC.

I came to the conclusion that we shouldn't do anything to manage toggle buttons, we already have what's necessary to manage toggle button in a a11y way.

Until the day comes where we are given clear design for toggle buttons for EVERY `intent` and `design` combination, we shouldn't consider implementing it as a true feature.

NOTE/SUMMARY:

If we DO want to implement it as a 1st class feature, it means we must require specs/figma for each `intent/design`combination first from designers. This is a prerequisite.

### Types of changes
- [x] 🧾 Documentation

### Screenshots - Animations

https://github.com/adevinta/spark/assets/2033710/e3baeb0e-7314-44f2-a8de-b7ed373fe343

### DIFF
--- a/packages/components/icon-button/src/IconButton.stories.tsx
+++ b/packages/components/icon-button/src/IconButton.stories.tsx
@@ -1,6 +1,8 @@
 import { StoryLabel } from '@docs/helpers/StoryLabel'
 import { Checkbox } from '@spark-ui/checkbox'
 import { Icon } from '@spark-ui/icon'
+import { LikeFill } from '@spark-ui/icons/dist/icons/LikeFill'
+import { LikeOutline } from '@spark-ui/icons/dist/icons/LikeOutline'
 import { Meta, StoryFn } from '@storybook/react'
 import { useState } from 'react'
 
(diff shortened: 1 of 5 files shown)


acd02:
LGTM @Powerplex  👍

packages/components/icon-button/src/IconButton.stories.tsx
```diff
@@ -128,3 +130,14 @@ export const Link: StoryFn = _args => (
     </IconButton>
   </div>
 )
+
+export const Toggle: StoryFn = () => {
+  const [pressed, setPressed] = useState(false)
+  const toggle = () => setPressed(!pressed)
+
+  return (
+    <IconButton aria-label="Add to favorites" aria-pressed={pressed} onClick={toggle}>
```
andresin87, line 139:
```suggestion
    <IconButton aria-label="Add to favorites" aria-pressed={pressed} onClick={toggle} design={pressed ? 'filled' : 'outlined'}>
```

Let me suggest to align the toggle demos aligned suggesting changing the design when toggling as an ideal situation. Changing outline-fill icons is not enough from my perspective.

andresin87 (approved):

PR #1510, opened 2023-09-29
feat(chip): add ref type to prop

tomasmax:
Add ref props type to Chip component to avoid Type error when passing reference prop from outside.

### Types of changes

- [x] 🧠 Refactor


andresin87:
thks

andresin87 (approved):

PR #1548, opened 2023-10-06
fix(chip): outline focuses

andresin87:
**TASK**: 

### Types of changes

- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [x] 💄 Styles

### DIFF
--- a/packages/components/chip/src/ChipClearButton.styles.tsx
+++ b/packages/components/chip/src/ChipClearButton.styles.tsx
@@ -55,8 +55,12 @@ export const chipClearButtonWrapperStyles = cva(
   }
 )
 
+// eslint-disable-next-line tailwindcss/no-custom-classname
 export const chipClearButtonStyles = cva(
-  ['rounded-full', 'focus-visible:outline-none focus-visible:u-ring-inset'],
+  [
+    'rounded-full text-[calc(var(--sz-6))] p-[var(--sz-2)] border-sm',
+    'focus-visible:outline-2 focus-visible:outline focus-visible:outline-offset-2  focus-visible:outline-[blue]',
+  ],
   {
     variants: {
       disabled: {
--- a/packages/components/chip/src/ChipClearButton.tsx
+++ b/packages/components/chip/src/ChipClearButton.tsx
@@ -1,5 +1,5 @@
 import { Icon } from '@spark-ui/icon'
-import { DeleteOutline } from '@spark-ui/icons/dist/icons/DeleteOutline'
+import { Close } from '@spark-ui/icons/dist/icons/Close'
 import React, { cloneElement, ComponentPropsWithoutRef, forwardRef, useCallback } from 'react'
 
 import {
@@ -19,8 +19,8 @@ export const ChipClearButton = forwardRef<HTMLSpanElement, ChipClearButtonProps>
   (
     {
       children = (
-        <Icon className="">
-          <DeleteOutline />
+        <Icon>
+          <Close />
         </Icon>
       ),
       tabIndex = 0,
@@ -41,7 +41,7 @@ export const ChipClearButton = forwardRef<HTMLSpanElement, ChipClearButtonProps>
     return (
       <span
         className={chipClearButtonWrapperStyles({
-          isBordered: design === 'dashed',
+          isBordered: ['outline', 'dashed'].includes(`${design}`),
           disabled: !!disabled,
           design,
         })}
--- a/packages/components/chip/src/variants/outlined.ts
+++ b/packages/components/chip/src/variants/outlined.ts
@@ -117,11 +117,11 @@ export const outlinedVariants = [
   {
     design: 'outlined',
     hasClearButton: false,
-    class: tw(['px-md']),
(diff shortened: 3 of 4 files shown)


packages/components/chip/src/ChipClearButton.tsx
```diff
@@ -41,7 +41,7 @@ export const ChipClearButton = forwardRef<HTMLSpanElement, ChipClearButtonProps>
     return (
       <span
         className={chipClearButtonWrapperStyles({
-          isBordered: design === 'dashed',
+          isBordered: ['outline', 'dashed'].includes(`${design}`),
```
andresz1, line 44: `includes(design)` instead ?

packages/components/chip/src/variants/outlined.ts
```diff
@@ -117,11 +117,11 @@ export const outlinedVariants = [
   {
     design: 'outlined',
     hasClearButton: false,
-    class: tw(['px-md']),
+    class: tw(['px-[calc(var(--spacing-md)-var(--border-width-sm))]']),
```
andresz1, line 120: Is is possible to use `calc` ? IMO this is not good in terms of performance :( because more classes will be added to the end css bundle and the size would be computed in runtime when there is no need

packages/components/chip/src/ChipClearButton.styles.tsx
```diff
@@ -55,8 +55,12 @@ export const chipClearButtonWrapperStyles = cva(
   }
 )
 
+// eslint-disable-next-line tailwindcss/no-custom-classname
 export const chipClearButtonStyles = cva(
-  ['rounded-full', 'focus-visible:outline-none focus-visible:u-ring-inset'],
+  [
+    'rounded-full text-[calc(var(--sz-6))] p-[var(--sz-2)] border-sm',
```
soykje, line 61: If I'm not wrong, you can directly consume your CSS variables:
```suggestion
    'rounded-full text-[--sz-6] p-[--sz-2] border-sm',
```

packages/components/chip/src/ChipClearButton.tsx
andresin87: design can be undefined and the typechecker gets wrong

packages/components/chip/src/variants/outlined.ts
andresin87: This class is aligned with the `dashed` design. Both designs shares the same classes so they are added nowadays.

packages/components/chip/src/ChipClearButton.styles.tsx
andresin87: thks

andresin87: omg. this doesnt-work 🤯

andresin87: I revert the changes. When applying `text-[--sz-6]` the result css applied is `color: var(--sz-6)` (0.325rem) which does nothing.

andresin87: @acd02 p[--sz-2] [font-size:--sz-8]

acd02 (approved):

packages/components/chip/src/ChipClearButton.styles.tsx
soykje: @acd02 To precise my initial remark: https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values. Official documentation says:
> When using a CSS variable as an arbitrary value, wrapping your variable in var(...) isn’t needed — just providing the actual variable name is enough: 
In this particular case it seems more related to the way `text-*` and `--sz-*` work together, IMHO :thinking:

acd02: @soykje , maybe that can help 👉 https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities

packages/components/chip/src/ChipClearButton.tsx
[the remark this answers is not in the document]
andresin87: i'll try, but typescript was returning an error

packages/components/chip/src/ChipClearButton.tsx
```diff
@@ -41,7 +41,7 @@ export const ChipClearButton = forwardRef<HTMLSpanElement, ChipClearButtonProps>
     return (
       <span
         className={chipClearButtonWrapperStyles({
-          isBordered: design === 'dashed',
+          isBordered: ['outline', 'dashed'].includes(design),
```
andresin87, line 44:
```suggestion
          isBordered: ['outline', 'dashed'].includes(`${design}`),
```

packages/components/chip/src/ChipClearButton.styles.tsx
soykje: Well let's leave @andresin87 with all this options to see what's better in his case :stuck_out_tongue: Loving tw simplicity... :innocent:

soykje (approved):

PR #1558, opened 2023-10-10
feat(divider): ready

andresin87:
**TASK**: 

🚨 Reviewed by @JeremiasUX and @elisa-hery 

### Types of changes


packages/components/divider/src/Divider.styles.ts
```diff
@@ -0,0 +1,97 @@
...
+      class: tw(['mx-lg min-h-sz-24 border-l-sm']),
+    },
+    {
+      isEmpty: false,
+      orientation: 'horizontal',
+      class: tw([
+        'flex-row my-sm grow-0',
+        'before:border-t-sm',
+        'after:border-t-sm',
+        '[&>*]:px-lg',
```
soykje, line 49: Why not having this style in `Divider.Content`?

packages/components/divider/src/DividerContent.tsx
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Change the component to the HTML tag or custom component of the only child.
+   */
+  asChild?: boolean
+  children?: ReactNode
+}
+
+export const DividerContent = forwardRef<HTMLSpanElement, DividerContentProps>(
+  ({ children, ...props }, ref) => {
+    return children ? (
```
soykje, line 13: Is there a real usecase of using `Divider.Content` without any children prop?

packages/components/divider/src/divider/intents.tsx
```diff
@@ -0,0 +1,26 @@
+import { tw } from '@spark-ui/internal-utils'
```
soykje, line 1: Shouldn't the directory be named `variants` instead of `divider`?

packages/components/divider/src/Divider.styles.ts
andresin87: no, cuz in that case we need to pass the prop to the Content with a Provider and it increases the logic for that "simple" line many more. Thats the simplest way

packages/components/divider/src/divider/intents.tsx
andresin87: you are right

packages/components/divider/src/DividerContent.tsx
andresin87: if someone do this

```tsx
const FOO = ({nomatherWhat}: {nomatherWhat?: string}) => {
	return (
      <>
		<h1>text</h1>
        <Divider>
          <Divider.Content>{nomatherWhat}</Divider.Content>
        <Divider>
		<p>Lorem ipsum dolor sit</p>
	 </>
}
```

User don't have to take care if `nomatherWhat` has a value or not to do not include non necessary html elements

packages/components/divider/src/Divider.doc.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+### Wrapped Content
+
+Users can also rotate the text to achieve the proper design.
+
+<Canvas of={stories.WrappedContent} />
+
+### Card Content
+
+<Canvas of={stories.Card} />
```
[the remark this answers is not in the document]
andresin87: It is taken form radix-ui

packages/components/divider/src/Divider.doc.mdx
```diff
@@ -0,0 +1,103 @@
...
+<Canvas of={stories.Default} />
+
+### Content
+
+Spark-ui gives you the opportunity
+to place an item on top of the divider by using its `Divider.Content` component as a children.
+
+<Canvas of={stories.Content} />
+
+### Children
```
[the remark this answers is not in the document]
andresin87: `asChild`

andresin87:
> I'm not entirely sure that we need this `Divider` component. I mean, if someone needs one, they could simply use the `hr` HTML tag element with a couple of Tailwind classes to style it the way they want, wdyt?

Radix had it and It is/was also defined in the roadmap when i started. At the same time, an hr will not be vertical aligned

packages/components/divider/src/Divider.doc.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Props
+
+<ExtendedArgTypes
+  of={Divider}
+  description="Visually or semantically separates content."
+  subcomponents={{
+    'Divider.Content': {
+      of: Divider.Content,
+      description: 'A label appearing on top of the divider if necessary.',
```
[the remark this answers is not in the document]
andresin87: it can place there whatever... icons, title elements, ...

packages/components/divider/src/DividerContent.tsx
andresin87: when there is no content provided the line is hanlded via 'before' & `after` with paddings so it that case it will have a line visually broken

packages/components/divider/src/Divider.styles.ts
andresin87: Yeap. Its 4 not having to drill down the orientation value from the divider to its content to indicate where the padding needs to be included. `asChild` + `children` combined make Provider the only way for sharing this setting between Content and its Parent Divider but it is too much complicated.

packages/components/divider/src/DividerContent.tsx
soykje: OK, but imho this kind of case should be only the responsibility of the dev, not ours... don't you think?

packages/components/divider/src/Divider.doc.mdx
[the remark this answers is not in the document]
soykje: Yep, but does it mean it is necessarily a good thing? :slightly_smiling_face:

packages/components/divider/src/Divider.doc.mdx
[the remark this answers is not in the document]
soykje: Yes, then the description should reflect that, don't you think? And not only referring to a "label", wdyt?

acd02:
@andresin87

>Radix had it and It is/was also defined in the roadmap when i started.
>At the same time:

>an hr will not be vertical aligned
>there is no way to add content inside an hr
>many other ds had it: (https://component.gallery/components/horizontal-rule/)

Then maybe we should follow [Spectrum](https://react-spectrum.adobe.com/react-spectrum/Divider.html)'s example on this one. Their approach seems more straightforward and simple, compared to implementations like [Decathlon](https://www.decathlon.design/726f8c765/p/28ad9b-divider/b/20e6b4) where they handle cases such as "**render a divider with text inside it**" which seem excessive imho

packages/components/divider/src/Divider.doc.mdx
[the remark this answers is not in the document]
andresin87: it also exist in next-ui

Powerplex (changes requested):
I am confused because I can't find specs for this component. I was expecting a simple divider but it comes with optional text on top, text positioning, orientation, etc.

Where does the need comes from ? How can we design review + check the specs ?

(I don't mind having it, I am just confused with our process)

Powerplex (approved):
Approving following our discussions, but I think next time we could decide to open as draft/proposal our experimental components as long as we don't have specs, so there is no fake sens of emergency in merging it.

I don't understand also why it went to design and QA validation before code review approval AND it being experimental.

I feel we can improve our communication here

acd02 (approved):

Issue #1533, opened 2023-10-05
Input/InputGroup (and its addons)


soykje:
For `focus-visible` (and sometimes `focus`) we are using box-shadows (Tailwind's `ring` API) to display an "outline". The issue is we also use `ring` to display what should be `borders` instead.

I suggest we check the following components 1 by 1 to see if we can replace they "neutral state" (non focused) styles to use proper borders and not box-shadows.

PR #1560, opened 2023-10-10
refactor(input): replaces ring by border

soykje:
**TASK**: #1533 

As statued in main issue we want to replace borders that have been previously achieved through `box-shadow` (aka `ring-*`) by using good ol' `border`.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
Capture d’écran du 2023-10-12 10-56-06
Capture d’écran du 2023-10-12 10-55-49


packages/components/input/src/Input.styles.ts
```diff
@@ -26,15 +26,10 @@ export const inputStyles = cva(
...
-        neutral: [
-          'ring-outline',
-          'hover:ring-outline-high',
-          'focus:ring-outline-high',
-          'disabled:ring-outline',
-        ],
-        success: ['ring-success', 'ring-success'],
-        alert: ['ring-alert', 'ring-alert'],
-        error: ['ring-error', 'ring-error'],
+        neutral: ['border-outline', 'hover:border-outline-high', 'disabled:border-outline'],
```
acd02, line 29: Is it normal that only the neutral intent has its own **disabled** styles?

soykje:
Not sure if this is what we want...
image

I'll look for updates from design + lead dev :+1:

soykje:
After discussions with both @JeremiasUX and @andresz1 we agreed that validation states (error, succes, alert) shouldn't occur at the same time as `disabled` state. It is all about user feedback, which shouldn't happen on disabled input... So small style reinforcements have been done "just in case" :+1: But no addition to stories about this very edge case, that could only happen on bad implementation.

packages/components/input/src/Input.styles.ts
soykje: See https://github.com/adevinta/spark/pull/1560#issuecomment-1757364903

soykje:
After last discussions, the focus styles for this component will only be black ones :+1:

acd02 (approved):

andresin87 (approved):

turolopezsanabria (approved):
👍🏼 Design check ok, thnx!

PR #1595, opened 2023-10-23
fix(input): set InputClearButton button type to "button"

warpdesign:
This avoids unwanted form submits.

## TYPE(SCOPE): TITLE

**TASK**: 

`InputGroup.ClearButton` renders a `button` element without setting its type, so the default behavior is `submit`: this means that clicking on clear button may submit the form.

Setting its type to `button` prevents unwanted form submission.

Insert an `InputGroup.ClearButton` inside a form: check that clicking on it submits the form.

Now set its button type to `button`: check that the form is not submitted.


soykje (approved):
Oh thx a lot! :+1:

andresz1 (approved):

Issue #1577, opened 2023-10-16
[Component] Rating


andresin87:
```[tasklist]
### Tasks
- [ ] https://github.com/adevinta/spark/pull/1605
- [ ] https://github.com/adevinta/spark/pull/1618
- [ ] https://github.com/adevinta/spark/pull/1642
- [ ] https://github.com/adevinta/spark/pull/1636
```

elisa-hery:
<google-sheets-html-origin>

Testing informations |  
-- | --
Does this ticket need to be tested ? | yes
Does this ticket need to be design reviewed ? | yes
Referent designer : | Laura
Spec : | [figma](https://www.figma.com/file/0QchRdipAVuvVoDfTjLrgQ/Spark-Specs-and-Handoff?type=design&node-id=7792-7244&mode=design&t=ZZFY1rwPpukfrZpN-0)
UATs : | [confluence](https://confluence.ets.mpi-internal.com/display/SPAR/Rating)
Snapshots scenarii : | [confluence](https://confluence.ets.mpi-internal.com/display/SPAR/Rating#Rating-snapshots)
demo app : | [confluence](https://confluence.ets.mpi-internal.com/display/SPAR/Rating#Rating-demo-app-configurator)

acd02:
👋 , hello @elisa-hery , actually, the Rating component has already been reviewed and (design) approved by @lauragautheron 👍

elisa-hery:
Thanks @acd02 , I know this one is already advanced, it was just a test for the "testing template"

soykje:
@elisa-hery Sorry for issue being closed after merging first of two PRs... I reopen it, here is the last one for testing purpose: https://github.com/adevinta/spark/pull/1618

elisa-hery:
@soykje @acd02 here are my comments : 

1. using **keyboard navigation** we can have a rate of 0/5 and it should not be possible

https://github.com/adevinta/spark/assets/142396647/05ab9699-2c30-4f73-a260-4c9c1979ecd7

2. **on mobile, for input rating**, when I click on a star, stars stay in the hover state. I have to click outside the rating component to make them back to selected state. Is it the expected behavior ?  See video below ⬇️ 

https://github.com/adevinta/spark/assets/142396647/b5b7b911-2faf-4811-90a1-56757a4d2e61

2. I've seen that you provide **3 sizes** but the specs mention "there is no need to block the size of the icons on the developer side". Is it on purpose ? Maybe you had to define sizes ?

[...]

PR #1605, opened 2023-10-26
feat(rating): Rating component API proposal

acd02:
**TASK**: #1577 

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [x] 🧪 Test
- [ ] 🧠 Refactor
- [x] 💄 Styles

### DIFF
--- a/packages/components/rating/src/Rating.test.tsx
+++ b/packages/components/rating/src/Rating.test.tsx
@@ -1,12 +1,96 @@
 import { render, screen } from '@testing-library/react'
-import { describe, expect, it } from 'vitest'
+import userEvent from '@testing-library/user-event'
+import { useState } from 'react'
+import { describe, expect, it, vi } from 'vitest'
 
 import { Rating } from './Rating'
 
+const utils = {
+  getInput: () =>
+    screen.getByRole('slider', {
+      name: /rating/i,
+    }),
+  getStars: (container: HTMLElement) =>
+    Array.from(container.querySelectorAll('[data-part="star"]')),
+}
+
 describe('Rating', () => {
   it('should render', () => {
-    render(<Rating />)
+    const { container } = render(<Rating defaultValue={2} />)
 
-    expect(screen.getByText(/rating/)).toBeInTheDocument()
+    expect(container.querySelector('[data-spark-component=rating]')).toBeInTheDocument()
+    expect(container.querySelectorAll('[data-part="star"]').length).toBe(5)
+  })
+
+  it('should be possible to interact', async () => {
+    const user = userEvent.setup()
+
+    const { container } = render(<Rating size="lg" defaultValue={1} />)
+
+    const stars = utils.getStars(container)
+    const input = utils.getInput()
+
+    // initial value
+    expect(input).toHaveValue('1')
+
+    await user.click(stars.at(2)!)
+    expect(input).toHaveValue('3')
+
+    await user.click(stars.at(0)!)
+    expect(input).toHaveValue('1')
+  })
+
+  it('should handle controlled mode', async () => {
+    function ControlledRating({ onValueChange }: { onValueChange: (value: number) => void }) {
+      const [value, setValue] = useState(2)
+
+      const handleInteraction = (value: number): void => {
+        onValueChange(value)
+        setValue(value)
+      }
+
+      return <Rating value={value} onValueChange={handleInteraction} />
+    }
+
+    const onValueChangeSpy = vi.fn()
+    const user = userEvent.setup()
+    const { container } = render(<ControlledRating onValueChange={onValueChangeSpy} />)
+
+    const stars = utils.getStars(container)
+    const input = utils.getInput()
+
+    // initial value
+    expect(input).toHaveValue('2')
+
+    await user.click(stars.at(4)!)
+    expect(input).toHaveValue('5')
+    expect(onValueChangeSpy).toHaveBeenCalledTimes(1)
+    expect(onValueChangeSpy).toHaveBeenCalledWith(5)
+  })
+
+  it('should not be possible to interact when in readOnly (in controlled mode)', async () => {
+    const user = userEvent.setup()
+    const handleValueChange = vi.fn()
+
+    render(<Rating value={1} onValueChange={handleValueChange} readOnly />)
+
+    const input = utils.getInput()
+
+    await user.click(input)
+    expect(handleValueChange).not.toHaveBeenCalled()
+  })
+
+  it('should not be possible to interact when disabled (in uncontrolled mode)', async () => {
+    const user = userEvent.setup()
+    const { container } = render(<Rating defaultValue={1} disabled />)
+
+    const stars = utils.getStars(container)
+    const input = utils.getInput()
+
+    expect(input).toBeDisabled()
+
+    // check that click doesn't do anything
+    await user.click(stars.at(4)!)
+    expect(input).toHaveValue('1')
   })
 })
--- a/packages/components/rating/src/Rating.test.tsx
+++ b/packages/components/rating/src/Rating.test.tsx
@@ -1,3 +1,4 @@
+import { fireEvent } from '@testing-library/dom'
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { useState } from 'react'
@@ -29,15 +30,24 @@ describe('Rating', () => {
 
     const stars = utils.getStars(container)
     const input = utils.getInput()
+    const firstStar = stars.at(0)
+    const thirdStar = stars.at(2)
+
+    if (!thirdStar || !firstStar) throw new Error('Stars element not found')
 
     // initial value
     expect(input).toHaveValue('1')
 
-    await user.click(stars.at(2)!)
+    await user.click(thirdStar)
     expect(input).toHaveValue('3')
 
-    await user.click(stars.at(0)!)
+    await user.click(firstStar)
     expect(input).toHaveValue('1')
+
+    // triggering a change event by directly interacting
+    // with the HTML input (using the keyboard for instance)
+    fireEvent.change(input, { target: { value: '5' } })
+    expect(input).toHaveValue('5')
   })
 
   it('should handle controlled mode', async () => {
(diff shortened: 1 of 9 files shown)


packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
-export const Rating = () => <>rating</>
+import { cx } from 'class-variance-authority'
+import {
+  type ChangeEvent,
+  ComponentPropsWithoutRef,
```
soykje, line 5:
```suggestion
  type ComponentPropsWithoutRef,
```

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
...
+  forwardRef,
+  type MouseEvent,
+  PropsWithChildren,
```
soykje, line 8:
```suggestion
  type PropsWithChildren,
```

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
...
+  /**
+   * When `true`, prevents the user from interacting.
+   * @default false
+   */
+  disabled?: boolean
+  /**
+   * Sets the size of the stars.
+   * @default 'md'
+   */
+  size?: Size
```
soykje, line 49: Why not doing as usual, and using the `RatingStarstylesProps` from styles? Just being curious... :sweat_smile:

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
...
+
+      if (isControlled) onValueChange?.(index + 1)
+      else setRatingValue(index + 1)
+
+      inputRef.current.focus()
+      inputRef.current.setAttribute('data-clicked', '')
+    }
+
+    function onInputChange(event: ChangeEvent<HTMLInputElement>) {
+      if (isNonInteractive) return
```
soykje, line 101: To simplify each function, wdyt of moving this logic such as:
```tsx
  {...!isNonInteractive && ({ onChange: onInputChange })}
```
? :thinking:

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
...
+
+      if (isControlled) onValueChange?.(Number(event.target.value))
+      else setRatingValue(Number(event.target.value))
+    }
+
+    function onStarMouseEnter(event: MouseEvent<HTMLDivElement>) {
+      if (isNonInteractive) return
+
+      const target = event.currentTarget
```
soykje, line 114: Wdyt of:
```tsx
  function onStarMouseEnter({ target }: MouseEvent<HTMLDivElement>) { // ...
```
to avoid this line? :smile:

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,182 @@
...
+      const [previousStars, followingStars] = splitAt(starRefList.current, currentStarIndex + 1)
+
+      previousStars.forEach(star => star.setAttribute('data-hovered', ''))
+      followingStars.forEach(star => star.removeAttribute('data-hovered'))
+    }
+
+    const handleStarRef = useCallback((elm: HTMLDivElement | null) => {
+      if (!elm) return
+      starRefList.current.push(elm)
+    }, [])
```
soykje, line 126: Shouldn't we add the `starRefList` in deps array?

packages/components/rating/src/RatingStar.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+const ratingStarStyles = cva(
+  ['peer', 'after:inset-none', 'group', 'relative', 'after:block after:absolute'],
+  {
+    variants: {
+      disabled: {
+        true: 'opacity-dim-3',
+        false: '',
+      },
+      readOnly: {
+        true: '',
```
soykje, line 14: Why having this variant if it has no styles? :thinking:

packages/components/rating/src/RatingStar.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+)
+
+const ratingStarIconStyles = cva('', {
+  variants: {
+    size: {
+      sm: 'text-caption-link',
+      md: 'text-body-1',
+      lg: 'text-display-1',
+    },
+    kind: {
```
soykje, line 44: For the `Button` we use `design` token (cf [here](https://sparkui.vercel.app/?path=/docs/components-button--docs#design))... Maybe we shoult reuse it, for consistency... wdyt?

packages/components/rating/src/RatingStar.tsx
```diff
@@ -0,0 +1,58 @@
...
+  ratingStarStyles,
+  type RatingStarstylesProps,
+} from './RatingStar.styles'
+import type { Size, StarValue } from './types'
+
+export interface RatingStarProps extends RatingStarstylesProps {
+  size?: Size
+  value: StarValue
+  onClick: (event: MouseEvent<HTMLDivElement>) => void
+  onMouseEnter: (event: MouseEvent<HTMLDivElement>) => void
```
soykje, line 18: These shouldn't be mandatory, as the component can be non interactive.

packages/components/rating/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export type Size = 'sm' | 'md' | 'lg'
```
soykje, line 1: Why wouldn't we reuse the `VariantProps` helper result to get this?

packages/components/rating/src/Rating.test.tsx
```diff
@@ -1,12 +1,127 @@
...
+    const user = userEvent.setup()
+
+    const { container } = render(<Rating size="lg" defaultValue={1} />)
+
+    const stars = utils.getStars(container)
+    const input = utils.getInput()
+    const firstStar = stars.at(0)
+    const thirdStar = stars.at(2)
+
+    if (!thirdStar || !firstStar) throw new Error('Stars element not found')
```
soykje, line 36: Why having this line in our UT? If L33 or L34 are falsy test would fail, and we also eventually could simply declare an expectation, wdyt? This syntax seems quite disturbing to me :thinking:

packages/components/rating/src/Rating.test.tsx
```diff
@@ -1,12 +1,127 @@
...
+
+    await user.click(thirdStar)
+    expect(input).toHaveValue('3')
+
+    await user.click(firstStar)
+    expect(input).toHaveValue('1')
+
+    // triggering a change event by directly interacting
+    // with the HTML input (using the keyboard for instance)
+    fireEvent.change(input, { target: { value: '5' } })
```
soykje, line 49: Can't you achieve the same goal with `user-event`?

packages/components/rating/src/Rating.test.tsx
```diff
@@ -1,12 +1,127 @@
...
+    const fifthStar = stars.at(4)
+
+    if (!fifthStar) throw new Error('Stars element not found')
+
+    // initial value
+    expect(input).toHaveValue('2')
+
+    await user.click(fifthStar)
+    expect(input).toHaveValue('5')
+    expect(onValueChangeSpy).toHaveBeenNthCalledWith(1, 5)
```
soykje, line 80: Why not simply:
```suggestion
    expect(onValueChangeSpy).toHaveBeenCalledWith(5)
```
?

packages/components/rating/src/Rating.tsx
acd02: Nope, that would be useless

packages/components/rating/src/Rating.tsx
acd02: Mostly because I found it more legible to declare the type this way 😄

packages/components/rating/src/Rating.tsx
acd02: Yes, I had that in mind initially. It's just that this syntax is ugly in my opinion. However, deep down, I find it more explicit to not call a function when a predicate is not met, rather than calling the function every time and checking inside

Will update it 👍

packages/components/rating/src/RatingStar.styles.ts
acd02: Because I use this variant inside the `compoundVariants`, to only apply the **cursor-pointer** styles when it's neither disabled nor in read-only mode

packages/components/rating/src/Rating.test.tsx
acd02: Nope, as we saw yesterday 😢
https://github.com/testing-library/user-event/issues/871

packages/components/rating/src/Rating.test.tsx
acd02: > If L33 or L34 are falsy test would fail

Unfortunately, no. Even adding tests that assert that the star is indeed there will not make TS infer that it is present in the following code. But, we could always resort to using the non-null assertion operator 🤷

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,180 @@
...
+            'flex',
+            'peer-focus-visible:u-ring peer-[[data-part=input][data-clicked]]:shadow-none'
+          )}
+        >
+          {Array.from({ length: 5 }).map((_, index) => (
+            <RatingStar
+              disabled={disabled}
+              readOnly={readOnly}
+              size={size}
+              onClick={() => isInteractive && onStarClick(index)}
```
andresin87, line 169: This isInteractive condition can be moved into onStarClickHandler in order to be able to be memoized

packages/components/rating/src/Rating.tsx
soykje: Got it, but we usually type (afaik) with `VariantProps` from cva in such cases... Nothing blocking for me, maybe someone else would have his 2cents to share with us? :slightly_smiling_face:

packages/components/rating/src/Rating.test.tsx
soykje: > To make sure we don't call the onValueChange prop more than we need to I guess
Then, I'm afraid there could be a misunderstanding here, as the `toHaveBeenNthCalledWith(a, b)` method test that the spied callback is  called with `b` value on `a` call.
If I understand correctly, you should here write something like:
```tsx
expect(onValueChangeSpy).toHaveBeenCalledTimes(1)
expect(onValueChangeSpy).toHaveBeenCalledWith(5)
```
Wdyt?

packages/components/rating/src/Rating.test.tsx
acd02: The purpose of this test is to make sure that we correctly linked the native `onChange` event with our internal React state. This ensures that when the user interacts with the component, the state is updated accordingly.

packages/components/rating/src/Rating.tsx
acd02: yes, but, I'm not so sure we need to memoize this function. According to the [React documentation](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere), you only really need to cache a function with `useCallback` in a few cases.

- Like if you're passing it as a prop to a component wrapped in `memo`, and you want to skip re-rendering if the value hasn't changed.
- Or if the function is later used as a dependency of some Hook, like if another function wrapped in `useCallback` depends on it

But in this case, it's neither of those. So, I think we can skip the memoization

soykje (approved):

acd02:
@andresin87 

>Is it experimental?
>Accessibility section needs to be included.
>It can be taken from https://component.gallery/components/rating/ (at the end) and also >https://baseweb.design/components/rating/#accessibility

Yes, the goal was to first have this PR approved and merged, and then make another one that adds all the stories 👍

soykje:
> @andresin87
> 
> > Is it experimental?
> > Accessibility section needs to be included.
> > It can be taken from https://component.gallery/components/rating/ (at the end) and also >https://baseweb.design/components/rating/#accessibility
> 
> Yes, the goal was to first have this PR approved and merged, and then make another one that adds all the stories 👍 (allowing Elisa to test it) Finally, a final pull request will be made to handle the publishing of the component.

Maybe we could update the PR title, and already prepare the final one, to make things clearer? :+1:

packages/components/rating/src/Rating.tsx
```diff
@@ -1 +1,179 @@
...
+        className="relative"
+        ref={ref}
+        data-spark-component="rating"
+        {...rest}
+        onMouseLeave={resetDataPartStarAttr}
+      >
+        <input
+          name={name}
+          id={id}
+          aria-label={rest['aria-label'] ?? 'rating'}
```
Powerplex, line 141: I wonder if we should do like IconButton and make the aria-label mandatory 🤔 

Here the risk is developers will never use it we should avoid to have hard-coded wordings in our components (this is part of our agreements).

For example `Dialog.CloseButton` does not have "close" as an aria-label by default.

WDYT ?

soykje: Completely agree! @acd02 do you want to take the point?

Issue #1600, opened 2023-10-25
[Tailwind] Update Spark Tailwind plugins to override "max-w-*" classes


acd02:
We should update our Tailwind plugins to override the `max-w-*` [classes](https://tailwindcss.com/docs/max-width) and other related classes from Tailwind.
This will allow us to handle the different base font-size conversion or remove those classes altogether

___

Here's an example to illustrate the problem:
Tailwind exposes this class → `max-w-md`  /* max-width: 28rem ===  448px */

However, we have a consumer (Polaris) that uses a base font-size of 10px, so they end up with:
`max-w-md`  /* max-width: 28rem ===  280px */

PR #1609, opened 2023-10-27
fix(tailwind-plugins): override "max-w-*" native tailwind classes

acd02:
**TASK**: #1600

Removing `max-w-*` native Tailwind classes from our Tailwind plugin because they are based on a fixed base font-size of 16px.

This lack of adaptability causes confusion at the call site, as the values displayed on the different TCV's or hinted by the IDE may differ from the actual computed values when using a different base font-size

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/utils/tailwind-plugins/spark-theme/index.js
+++ b/packages/utils/tailwind-plugins/spark-theme/index.js
@@ -86,6 +86,17 @@ module.exports = plugin.withOptions(
 
     if (!themes.default) throw new Error(missingDefaultThemeErrorMsg)
 
-    return { theme: getCSSVariableReferences(themes.default) }
+    return {
+      theme: {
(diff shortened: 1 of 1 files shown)


Powerplex (approved):

andresin87 (approved):

acd02:
@andresin87 
> Why only max-w- and not min-w- ?
Because Tailwind only add these extra values for the max-width classes, see 👉 https://sourcegraph.com/github.com/tailwindlabs/tailwindcss-from-zero-to-production/-/blob/08-optimizing-for-production/tailwind-full.config.js?L453

PR #1618, opened 2023-11-02
docs(rating): component rating docs

soykje:
**TASK**: #1577 

Following [API proposal](https://github.com/adevinta/spark/pull/1605), we now want to complete documentation on Rating component.

### Types of changes
- [x] 🧾 Documentation
- [x] 📷 Demo

### Screenshots - Animations
image

### DIFF
--- a/packages/components/rating/src/Rating.doc.mdx
+++ b/packages/components/rating/src/Rating.doc.mdx
@@ -54,6 +54,19 @@ Use `size` prop to set the size of the rating control stars.
 
 <Canvas of={stories.Size} />
 
+## Advanced usage
+
+### Rounded values
+
+The **Rating** component includes a simple rounding mechanism to guarantee a nice display.
+
+- On **input mode** it will use a step of **1** (for accessibility concerns)
+- On **read-only mode** it will use a **0.5** step. In this last case the 0.5 step should be considered as a "visible" one, as we round using a 0.25 step behind the scene. This allows us to define more precisely if a value should be rounded down or up.
+
+<Canvas of={stories.Rounded} />
+
+If for any reason you need some specific rounding mechanisms, feel free to define yours and provide your rounded value.
+
 ## Accessibility
 
 Follows WCAG 2.0 standards for keyboard accessibility guidelines.
--- a/packages/components/rating/src/Rating.doc.mdx
+++ b/packages/components/rating/src/Rating.doc.mdx
@@ -1,5 +1,6 @@
 import { Meta, Canvas } from '@storybook/addon-docs'
-import { ArgTypes } from '@storybook/blocks';
+import { ArgTypes } from '@storybook/blocks'
+import { Kbd } from '@spark-ui/kbd'
 
 import { Rating } from '.'
 
@@ -9,7 +10,7 @@ import * as stories from './Rating.stories'
 
 # Rating
 
-Ratings let users see and/or set a star rating for a product or other item.
+Ratings lets users see and/or set a star rating for a product or other item.
 
 ## Install
 
@@ -20,7 +21,7 @@ npm install @spark-ui/rating
 ## Import
 
 ```tsx
-import { Rating } from "@spark-ui/rating"
+import { Rating } from '@spark-ui/rating'
 ```
 
 ## Props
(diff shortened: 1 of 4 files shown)


packages/components/rating/src/Rating.doc.mdx
```diff
@@ -30,4 +31,36 @@ import { Rating } from "@spark-ui/rating"
...
+
+### Sizes
+
+Use `size` prop to set the size of the rating control stars.
+
+<Canvas of={stories.Size} />
+
+## Accessibility
+
+Follows WCAG 2.0 standards for keyboard accessibility guidelines.
```
Powerplex, line 59: I can't find the WCAG link for the keyboard navigation standard of a "Rating", maybe we can add a link to this one instead ? https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/

packages/components/rating/src/Rating.stories.tsx
```diff
@@ -9,4 +10,21 @@ const meta: Meta<typeof Rating> = {
 
 export default meta
 
-export const Default: StoryFn = _args => <Rating />
+const sizes: RatingProps['size'][] = ['sm', 'md', 'lg']
+
+export const Default: StoryFn = _args => <Rating defaultValue={3} />
+
+export const Readonly: StoryFn = _args => <Rating defaultValue={3.5} readOnly />
```
Powerplex, line 17: Maybe for this story we could set "3.6" or "3.62" to show that it rounded to the closest half star value.

Powerplex (approved):

andresin87 (approved):

packages/components/rating/src/Rating.stories.tsx
andresin87: Some nice a11y definitions:
- A the end of the page: https://component.gallery/components/rating/
- https://baseweb.design/components/rating/

packages/components/rating/src/Rating.doc.mdx
soykje: After some researches, we noticed there were two patterns when implementing a rating component: using a slider input, or a group of radio inputs. With @acd02 we chose the first option, that seemed more appropriate to us (and following the Adobe Spectrum implementation). The pattern you linked only fit with second choice... :sweat_smile:

packages/components/rating/src/Rating.stories.tsx
soykje: @andresin87 Not sure to get your point... Is there any improvement you want to suggest on current a11y definitions here? :thinking: 
@Powerplex I don't have any strong opinion on that, but I wanted to keep thins simple. Also to avoid people thinking we will manage the rounding logic... In our implementation we only "round" to preserve the displayed value... wdyt?

packages/components/rating/src/Rating.tsx
```diff
@@ -129,7 +129,7 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
 
     return (
       <div
-        className="relative"
+        className="relative inline-flex"
```
acd02, line 132: Why adding this class?

soykje: As we focus the group, we didn't want to end with a full-width block (thx to the `div`) as our stars only fill a part of it...
Before: image
After: 
image

acd02: I see. Btw, we could achieve the same result by using `w-max` here. I have nothing against using flex here, but since this element only has one true child, we don't take full advantage of the flex layout. Either way works for me 🤷

andresin87 (approved):

kikoruiz (approved):

Powerplex:
@andresin87 regarding the issue of the input having `value="3"` on the `readOnly` story. I read the spec of the range input, and it seems `value` here is a `string representation of the selected number`. We don't have to round the number to the closest integer it seems.

I think @acd02 you checked on a screen reader and it reads out loud the `value` attribute. 

So maybe we could just set the `value` attribute to `3.5` and it would work well. Screen readers users would have the same value as other users. WDYT ?

soykje:
> @andresin87 regarding the issue of the input having `value="3"` on the `readOnly` story. I read the spec of the range input, and it seems `value` here is a `string representation of the selected number`. We don't have to round the number to the closest integer it seems.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#value
> 
> I think @acd02 you checked on a screen reader and it reads out loud the `value` attribute.
> 
> So maybe we could just set the `value` attribute to `3.5` and it would work well. Screen readers users would have the same value as other users. WDYT ?
> 
> Our rating component would look like this (my proposal is to keep the `step` to `0.5` as our component support half-stars, and not open the prop):
> 
> ```
> <input type="range" min="0" max="5" step="0.5" value="3.5" />
> ```
> 
> Capture d’écran 2023-11-09 à 15 26 09

So in short terms you suggest to udpate the input step to `0.5`, right? Just to be sure I understood correctly :sweat_smile:

Powerplex:
@acd02 and these are the spec when the input is interactive. In `readOnly` mode it will display half-stars.

acd02:
I believe that this method of discussing is not the most practical, @Powerplex . Additionally, the goal of this PR is to add stories, which is not directly related to the internals of the component. Maybe you could create a separate github discussion for this?

soykje:
@andresin87 @Powerplex @acd02 Here is a proposal PR: https://github.com/adevinta/spark/pull/1636

PR #1636, opened 2023-11-09
feat(rating): upd rating input step

soykje:
**TASK**: #1577 

Regarding @andresin87 and @Powerplex remarks, we may want to update the Rating input value, for accessibility reasons.

### Types of changes
- [x] 🧠 Refactor

### DIFF
--- a/packages/components/rating/src/Rating.tsx
+++ b/packages/components/rating/src/Rating.tsx
@@ -11,7 +11,7 @@ import {
 } from 'react'
 
 import { RatingStar, type RatingStarProps } from './RatingStar'
-import { getNearestDecimal, getStarValue, splitAt } from './utils'
+import { getNearestHalfDecimal, getStarValue, splitAt } from './utils'
 
 export interface RatingProps extends PropsWithChildren<ComponentPropsWithoutRef<'div'>> {
   /**
@@ -147,10 +147,10 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
           type="range"
           min="0"
           max="5"
-          step={1}
+          step={readOnly ? 0.5 : 1}
           disabled={disabled}
           readOnly={readOnly}
-          value={getNearestDecimal(value)}
+          value={getNearestHalfDecimal(value ?? 0)}
           onChange={event => isInteractive && onInputChange(event)}
           onBlur={resetDataPartInputAttr}
         />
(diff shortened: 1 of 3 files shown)


packages/components/rating/src/Rating.tsx
```diff
@@ -145,10 +145,10 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
           type="range"
           min="0"
           max="5"
-          step={1}
+          step={readOnly ? 0.5 : 1}
```
acd02, line 148:
```suggestion
          step={1}
```

packages/components/rating/src/Rating.tsx
```diff
@@ -145,10 +145,10 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
...
           disabled={disabled}
           readOnly={readOnly}
-          value={getNearestDecimal(value)}
+          value={getNearestHalfDecimal(value ?? 0)}
```
acd02, line 151:
```suggestion
          value={isInteractive ? getNearestHalfDecimal(value ?? 0) : value}
```

packages/components/rating/src/Rating.tsx
soykje: Sorry, but I don't understand your remark as it seems to be the opposite from what we discussed altogether yesterday :thinking:

acd02: imho there is no point in updating the step, knowing that you cannot interact with the component in readOnly mode

acd02: To me, the point was that we wanted to avoid rounding the value of the input (the value that is announced through voice over) when in non-interactive mode

soykje: Just took the time to investigate on this, and it appears that the voice syntax (at least on Linux?) doesn't simply read the input value. He also seems to take into account the `step`, and in your case doesn't give the expected result. Happy to discuss, if some of you can also check with their voice syntax tool :+1:

acd02: I'll take a look this afternoon or tomorrow 👍

Powerplex (approved):

packages/components/rating/src/Rating.tsx
acd02: After testing it with voice over, you're right, we do need to update the **step** value. So, this PR looks good to me 👍

acd02 (approved):

PR #1642, opened 2023-11-13
feat(rating): prevent null value on keyboard interaction

soykje:
**TASK**: #1577 

As user can't go back to zero when updating `Rating` component's value through mouse/pointer interactions, we want the same behaviour for keyboard.

### Types of changes
- [x] 🧠 Refactor

### DIFF
--- a/packages/components/rating/src/Rating.tsx
+++ b/packages/components/rating/src/Rating.tsx
@@ -97,9 +97,11 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
     }
 
     function onInputChange(event: ChangeEvent<HTMLInputElement>) {
-      // avoiding unnecessary calls to the onValueChange prop
-      // when the value remains unchanged
-      if (valueRef.current === Number(event.target.value)) return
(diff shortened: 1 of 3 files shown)


packages/components/rating/src/Rating.tsx
```diff
@@ -105,6 +106,16 @@ export const Rating = forwardRef<HTMLDivElement, RatingProps>(
       setRatingValue(Number(event.target.value))
     }
 
+    function onInpuKeyDown(event: KeyboardEvent<HTMLInputElement>) {
```
acd02, line 109: Instead of adding this new handler, I think we could just bail out from the `onInputChange` if `event.target.value === '0'`

soykje: Not suze, as we need to still accept a 0 value at the beginning. We just want to prevent user from decreasing value to 0.

acd02: Did you try it? 😉

soykje: I hate when you're right... :disappointed: Thx :+1:

PR #1650, opened 2023-11-17
chore(rating): publish component

soykje:


acd02:
@soykje , before publishing, shouldn't we take into account Laura's comment?
https://github.com/adevinta/spark/issues/1577#issuecomment-1815971608

PR #1664, opened 2023-11-21
feat(popover): add inset prop to popover

soykje:
**TASK**: #1653 

There is a common use case in Polaris where users want to have elements within the Popover component that takes up the full width.

Currently, achieving this requires overriding the internal padding of the component. To prop would make this use case easier and more intuitive.

When set to true, this prop will remove the internal padding of the component, allowing for a full-width image or content.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
image

### DIFF
--- a/packages/components/popover/src/Popover.doc.mdx
+++ b/packages/components/popover/src/Popover.doc.mdx
@@ -76,6 +76,12 @@ The content is the popover itself.
 
 <Canvas of={stories.Default} />
 
+### Inset
+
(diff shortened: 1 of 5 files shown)


acd02 (approved):

turolopezsanabria (approved):
👍🏼

Issue #1398, opened 2023-09-06
[Documentation] - Adding jsdoc descriptions to spark packages


Powerplex:
Adding jsdoc comment to describe every component's props would automatically add those description to the docgen.json published with each package.

We could then use this data on the new documentation website to display description that comes directly from our packages, instead of hardcoding in the MDX

PR #1704, opened 2023-11-28
docs: add missing comments to props (types)

kikoruiz:
## docs: add missing comments to props (types)

**TASK**: #1398 

Adding jsdoc comment to describe every component's props would automatically add those description to the docgen.json published with each package.

We could then use this data on the new documentation website to display description that comes directly from our packages, instead of hardcoding in the MDX

### Types of changes
- [x] 🧾 Documentation


packages/components/dialog/src/Dialog.tsx
```diff
@@ -3,7 +3,28 @@ import { type ReactElement } from 'react'
 
 import { DialogProvider } from './DialogContext'
 
-export type DialogProps = RadixDialog.DialogProps
+export interface DialogProps {
```
kikoruiz, line 6: Not sure if we should do this. It's the only way to register the comment for each type we use from an external resource or library, but this way we're forced to rewrite the entire interface 😞

Powerplex: I think we should. Only because Radix does not commit to providing JSDoc for their components props.

Their documentation is using static documentation and no JSDoc like us, so basically they write the descriptions twice. I think it's the only way for us to not be dependant on how they write/expose their component props.

kikoruiz: Thanks for your feedback!! Left it as it is, I only had to fix some type errors

andresin87 (approved):

packages/components/input/src/Input.styles.ts
```diff
@@ -35,19 +41,34 @@ export const inputStyles = cva(
...
+      /**
+       * Sets if there is an addon before the input text.
+       */
       hasLeadingAddon: {
         true: ['rounded-l-none'],
         false: ['rounded-l-lg'],
       },
+      /**
+       * Sets if there is an addon after the input text.
+       */
```
acd02, line 53: I don't think this type of property is exposed to our consumers, right?
same goes for the following `has*` props below

acd02 (approved):

packages/components/input/src/Input.styles.ts
kikoruiz: Yep, these comments are not exposed into the compiled files and they will only appear in the `docgen.json`

PR #1708, opened 2023-11-28
1684 components dropdown controlled mode

Powerplex:
## TYPE(SCOPE): TITLE

**TASK**: #1684 

Controlled mode is provided for two features:
- `value` management using `value, onValueChange, defaultValue` to stick with other Spark components API
- `open` management using `open, onOpenChange, defaultOpen` to stick with other Spark components API

### Types of changes

- [x] 🧾 Documentation

### DIFF
--- a/packages/components/dropdown/src/Dropdown.stories.tsx
+++ b/packages/components/dropdown/src/Dropdown.stories.tsx
@@ -1,3 +1,4 @@
+import { Button } from '@spark-ui/button'
 import { FormField } from '@spark-ui/form-field'
 import { Tag } from '@spark-ui/tag'
 import { Meta, StoryFn } from '@storybook/react'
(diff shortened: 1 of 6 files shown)


packages/components/dropdown/src/Dropdown.stories.tsx
```diff
@@ -34,6 +36,67 @@ export const Default: StoryFn = _args => {
   )
 }
 
+export const Controlled: StoryFn = () => {
+  const [value, setValue] = useState('book-1')
+
+  return (
+    <div className="w-sz-480 pb-[300px]">
+      <Dropdown value={value} onValueChange={setValue}>
```
andresin87, line 44: The value is not handled by the Dropdown.value component?

Powerplex: Actually no :) We read the Radix doc too fast last time. `Dropdown.Value` is just for the display of the current selected value inside the trigger, but it does not manage it.

All the main props are on the root element: https://www.radix-ui.com/primitives/docs/components/select#root

And this is also how we did for other components until now:

PR #1709, opened 2023-11-29
feat(dropdown): add leadingIcon

andresin87:
**TASK**: 

### Types of changes

- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/dropdown/src/utils.ts
+++ b/packages/components/dropdown/src/utils.ts
@@ -1,4 +1,4 @@
-import React, { type FC, isValidElement, type ReactElement, type ReactNode } from 'react'
+import React, { Children, type FC, isValidElement, type ReactElement, type ReactNode } from 'react'
 
 import { type ItemProps } from './DropdownItem'
 import { type DropdownItem, type ItemsMap } from './types'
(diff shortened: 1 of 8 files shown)


Powerplex (approved):

packages/components/dropdown/src/utils.ts
```diff
@@ -35,6 +35,18 @@ const getElementId = (element?: ReactElement) => {
   return element ? (element.type as FC & { id?: string }).id : ''
 }
 
+export const findElement =
```
kikoruiz, line 38: I think we should have some kind of generic "utils" package in order to consume it from any component, instead of creating this functions locally, WDYT?

kikoruiz (approved):

packages/components/dropdown/src/utils.ts
andresin87: yes, I agree. Its repeated in many places

andresin87: - chip
- select
- dropdown
- input (not exactly the same function but this can work too for its purpose)

Issue #1581, opened 2023-10-16
[Component] Progress tracker


thomas-leguellec:
```[tasklist]
### Tasks
- [ ] https://github.com/adevinta/spark/issues/1696
- [ ] https://github.com/adevinta/spark/pull/1714
- [ ] https://github.com/adevinta/spark/pull/1758
- [ ] https://github.com/adevinta/spark/pull/1765
```

elisa-hery:
<google-sheets-html-origin>

Testing informations |  
-- | --
Risk level ? | moderate
Does this ticket need to be design reviewed ? | yes
Referent designer : | @JeremiasUX 
Spec : | [figma](https://www.figma.com/file/0QchRdipAVuvVoDfTjLrgQ/Spark-Specs-and-Handoff?type=design&node-id=12212-6713&mode=design&t=FvppNKarxpA4HxdE-0)
UATs : | [confluence](https://confluence.ets.mpi-internal.com/display/SPAR/Progress+tracker)
configurator : | [confluence](https://confluence.ets.mpi-internal.com/display/SPAR/Progress+tracker#Progress+tracker-demo-app-configurator)

soykje:
@elisa-hery @JeremiasUX Here is the testing URL : https://spark-40nikjmyx-spark-team.vercel.app/?path=/docs/experimental-progresstracker--docs

elisa-hery:
@soykje I've got some comments / questions ! 

1. I noticed a bug when clicking from step 2 to step 1, the steps are going crazy (see video) 
https://github.com/adevinta/spark/assets/142396647/b9ad6f2a-90db-4161-ab23-00d7506a3ec7

2. I don't understand the state of progress indicator when you can hover them, they are focusable, but not selectable
https://github.com/adevinta/spark/assets/142396647/41c2ae6a-4180-40c0-952e-70980837d7cd

3. vertical alignement between text & progress indicator does not seem same as in the specs ⬇️ 
![Capture d’écran 2024-01-04 à 18 19 30](https://github.com/adevinta/spark/assets/142396647/e7b585c1-dc6d-4bc5-8995-2253acb6a5d4)


4. for "long" progress label, text should go next line but it doesn't (see specs ⬇️ ), we could see with designers if it matters for V1 or not
![Capture d’écran 2024-01-04 à 18 31 10](https://github.com/adevinta/spark/assets/142396647/9917d018-45f5-4735-b0fd-9b351da07061)

[...]

PR #1714, opened 2023-12-01
feat(progress-tracker): add base markup for progress-tracker

soykje:
**TASK**: #1581 

Getting started with component, with base features.

**NB:** Interactivity behaviour will be updated/refactored after specs updates, in further PR.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 💄 Styles


packages/components/progress-tracker/src/ProgressTrackerStep.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const stepId = `step-${ID}`
+    const stepIndex = [...steps.keys()].indexOf(stepId)
+
+    const disabledAfter = useMemo(() => {
+      const nextStepId = [...steps.keys()][stepIndex + 1]
+
+      return !!(nextStepId && steps.get(nextStepId)?.includes('disabled'))
+    }, [steps, stepIndex])
+
+    const [progressState, setProgressState] =
```
acd02, line 52: I think we can do without this `useState`, i.e.:
```tsx
progressState = (() => {
  if (stepIndex === activeStepIndex) 'active'
  else if (stepIndex < activeStepIndex) return 'complete'
  return 'incomplete'
})()
```

soykje: Thx, I'll test this option :+1: But to be curious, what would be the real gain of this method?

acd02: https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations

I would flip this question around, and ask you what's the benefit of using `useState` and `useEffect` when you can get by without them 😉

soykje: I knew you would send this link...

packages/components/progress-tracker/src/ProgressTracker.tsx
```diff
@@ -1,9 +1,84 @@
...
+  /**
+   * The size of the step indicators
+   * @default 'lg'
+   */
+  size?: 'sm' | 'md' | 'lg'
+  /**
+   * The index of the current step.
+   * @default 0
+   */
+  stepIndex?: number
```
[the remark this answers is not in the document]
soykje: I don't understand why we should change now, as it has been previously discussed without any remarks on this naming :thinking:
Anyway, I think we should keep the `*Index` information, as prop expects... an index :sweat_smile: Just took a look on some other DS like [Chakra](https://chakra-ui.com/docs/components/stepper/props#props) and not sure the `active` or `current` is more relevant here... But I don't have a really strong opinion on it^^ Happy to discuss.

packages/components/progress-tracker/src/ProgressTrackerStep.tsx
```diff
@@ -0,0 +1,117 @@
...
+      setSteps,
+      size,
+      readOnly,
+    } = useProgressTrackerContext()
+
+    const ID = useId()
+    const stepId = `step-${ID}`
+    const stepIndex = [...steps.keys()].indexOf(stepId)
+
+    const disabledAfter = useMemo(() => {
```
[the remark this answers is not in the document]
soykje: You're right, as deps come from props/context, reconciliation already works :+1:

Powerplex (approved):

acd02 (approved):

andresin87:
It looks like there is a problem with the interaction on this PR

https://github.com/adevinta/spark/assets/1674036/c7eea564-2cae-448c-a2b6-e9aaeeb5d480

We need to review at the end of all the PRs

andresin87:
progress-tracker is using basic, main, support naming convention and this is only availabe using
"@spark-ui/theme-utils": "^4.0.0",
as a peer-dependency. Please do not forget

andresin87 (approved):

PR #1744, opened 2023-12-13
fix(switch): improved animation performance

Powerplex:

To discuss. Some teams using our Switch component reported that it's toggle animation was lagging when clicking on it trigger heavy computation of their pages.

Heavy computation is an issue on their side, but it seems those styles help makes seems smoother (we tested it).

WDYT ? 

### Types of changes

- [x] 💄 Styles

### DIFF
--- a/packages/components/switch/src/SwitchInput.styles.ts
+++ b/packages/components/switch/src/SwitchInput.styles.ts
@@ -3,12 +3,12 @@ import { cva, VariantProps } from 'class-variance-authority'
 
 export const styles = cva(
   tw([
-    'group relative inline-flex flex-shrink-0 items-center self-baseline',
+    'relative flex-shrink-0 self-baseline',
     'cursor-pointer',
     'rounded-full border-transparent',
     'hover:ring-4',
     'transition-colors duration-200 ease-in-out',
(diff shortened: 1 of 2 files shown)


acd02:
@Powerplex ,

Did you try using the `will-change` property? Because it seems from your updates that we will be animating the `margin` instead of using transforms, which is counterintuitive, as it is [widely known](https://web.dev/articles/animations-guide) that one should always prefer using **transform** for performant CSS animations.

packages/components/switch/src/SwitchInput.styles.ts
```diff
@@ -49,13 +49,25 @@ export const styles = cva(
...
+    },
+  }
+)
+
 export const thumbStyles = cva(
   [
-    'absolute flex items-center justify-center',
+    'absolute left-none top-none flex items-center justify-center',
     'bg-surface',
-    'group-spark-state-checked:left-full group-spark-state-checked:-translate-x-full',
```
Powerplex, line 56: The perf improvement comes from here, as explained in [this article](https://codeburst.io/low-performance-css-transitions-and-some-alternatives-f84ff35bf07) shared by Marius.

Layout properties such as `left` have bad performance when animated. The trick is to only use translateX if possible.

soykje (approved):

PR #1809, opened 2024-01-11
fix(dropdown): z-index and focus behaviour issues

Powerplex:

1. fixed bug related to z-index inside popover. Downshift refs should always be merged with our forwarded refs.
2. fixed bug related to focus forwarding. When Dropdown was used inside a Dialog, it was breaking `onOpenAutoFocus` behaviour.

### Types of changes

### DIFF
--- a/packages/components/dropdown/src/DropdownItem.tsx
+++ b/packages/components/dropdown/src/DropdownItem.tsx
@@ -1,10 +1,11 @@
+import { useMergeRefs } from '@spark-ui/use-merge-refs'
 import { cva, cx } from 'class-variance-authority'
-import { forwardRef, ReactNode, type Ref } from 'react'
+import { forwardRef, HTMLAttributes, ReactNode, type Ref } from 'react'
 
 import { useDropdownContext } from './DropdownContext'
 import { DropdownItemProvider, useDropdownItemContext } from './DropdownItemContext'
 
-export interface ItemProps {
+export interface ItemProps extends HTMLAttributes<HTMLLIElement> {
   disabled?: boolean
   value: string
   children: ReactNode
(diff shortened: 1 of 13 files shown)


packages/components/dropdown/src/DropdownItem.tsx
```diff
@@ -1,10 +1,11 @@
+import { useMergeRefs } from '@spark-ui/use-merge-refs'
 import { cva, cx } from 'class-variance-authority'
-import { forwardRef, ReactNode, type Ref } from 'react'
+import { forwardRef, HTMLAttributes, ReactNode, type Ref } from 'react'
```
soykje, line 3: For consistency?
```suggestion
import { forwardRef, type HTMLAttributes, type ReactNode, type Ref } from 'react'
```

packages/components/dropdown/src/DropdownPopover.tsx
```diff
@@ -10,27 +10,26 @@ export const Popover = forwardRef(
...
+    const { isOpen, setHasPopover } = useDropdownContext()
 
     useEffect(() => {
       setHasPopover(true)
 
       return () => setHasPopover(false)
-    }, [setHasPopover])
-
-    if (!hasPopover) return children
+    }, [])
```
soykje, line 24: I guess you should add an ignore here, for empty deps array... :thinking:

packages/components/dropdown/src/utils.ts
```diff
@@ -93,3 +93,17 @@ export const getItemsFromChildren = (children: ReactNode): ItemsMap => {
 
   return newMap
 }
+
+export const hasChildComponent = (children: ReactNode, displayName: string): boolean => {
```
soykje, line 97: As discussed, we may want to gather as web team, to define a common strategy about this kind of stuff... :+1:

soykje (approved):
:fire:

PR #1843, opened 2024-01-25
fix(dropdown): fixed missing ItemIndicator on Safari

Powerplex:
**TASK**: #1844 

During testing session, we found out that ItemIndicator icon was invisible on safari only.

We forgot to wrap the Check icon with `@spark-ui/icon`. I don't know why it was only broken on Safari, but it fixes the bug.

### Types of changes

### Screenshots - Animations

Capture d’écran 2024-01-25 à 16 12 56

### DIFF
--- a/packages/components/dropdown/src/DropdownItemIndicator.tsx
+++ b/packages/components/dropdown/src/DropdownItemIndicator.tsx
@@ -1,8 +1,8 @@
+import { Icon } from '@spark-ui/icon'
 import { Check } from '@spark-ui/icons/dist/icons/Check'
 import { cx } from 'class-variance-authority'
 import { forwardRef, ReactNode, type Ref } from 'react'
 
-import { useDropdownContext } from './DropdownContext'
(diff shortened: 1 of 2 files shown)


andresin87 (approved):

packages/components/dropdown/src/DropdownItemIndicator.tsx
```diff
@@ -14,9 +14,12 @@ export interface ItemIndicatorProps {
 export const ItemIndicator = forwardRef(
   ({ className, children, label }: ItemIndicatorProps, forwardedRef: Ref<HTMLSpanElement>) => {
```
acd02, line 15:
```suggestion
  ({ className, label ,children =  <Icon size="sm"><Check aria-label={label} /></Icon> }: ItemIndicatorProps, forwardedRef: Ref<HTMLSpanElement>) => {
```

acd02 (approved):

PR #1659, opened 2023-11-20
feat(alert-dialog): release first version

andresz1:
## alert-dialog

**TASK**: None

### Description, Motivation and Context
This PR will release the first version of the `AlertDialog`

PR #1873, opened 2024-02-06
fix(dropdown): release dropdown and select as stable

Powerplex:

- Releasing `@spark-ui/dropdown` as stable
- Releasing `@spark-ui/select` as stable

### Types of changes

### DIFF
--- a/packages/components/dropdown/package.json
+++ b/packages/components/dropdown/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@spark-ui/dropdown",
-  "version": "0.21.5",
+  "version": "1.0.0-0",
   "description": "Displays a list of options for the user to pick from—triggered by a button. Differs from Select in that it offers multiple select and its list is not native.",
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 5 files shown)


soykje (approved):
:tada:

packages/components/dropdown/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@spark-ui/dropdown",
-  "version": "0.21.5",
+  "version": "1.0.0",
```
andresz1, line 3: I think that if you do it like this, you will publish `1.1.0`. Perhaps you should do `1.0.0-0`

andresz1: See [example](https://github.com/adevinta/spark/pull/1659/files#diff-b023b4d5257c64b23cb36f5d36594f7e6c0d80deca00ee635f5a69a86163e187R3)

acd02 (approved):

packages/components/dropdown/package.json
Powerplex: It would publish 1.0.1 not 1.1.0 because I declared it as a fix but you are right :)

I didn't know this trick.

Issue #1948, opened 2024-03-11
[Component] - Dialog - apply padding right to Dialog.Title when Dialog.CloseButton is present


acd02:
We need to add functionality to detect the presence of the CloseButton within the dialog and automatically apply padding right to the dialog's title.
This improvement will prevent the need for consumers of the component to manually add this padding at the call site

PR #1949, opened 2024-03-11
feat(dialog): apply padding right to dialog title when closebutton is present

acd02:
**TASK**: #1948

This PR ensures that when we detect the presence of the CloseButton within the dialog, we automatically apply padding right to the dialog's title.

This improvement will prevent the need for consumers of the component to manually add this padding at the call site.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2400,6 +2400,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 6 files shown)


packages/components/dialog/src/DialogContent.tsx
```diff
@@ -13,10 +21,26 @@ export interface ContentProps extends RadixDialog.DialogContentProps, DialogCont
...
+
+    const findElement = (...values: string[]) => {
+      const children = Children.toArray(childrenProp).filter(isValidElement)
+
+      return children.find(child => values.includes(getElementId(child) || ''))
+    }
+
+    const closeButton = findElement('CloseButton')
+
+    const { setIsFullScreen, setHasCloseButton } = useDialog()
```
Powerplex, line 39: I think all the added logic here should be inside `useDialog` instead (the purpose of this hook was to centralize the dialog logic)

acd02: Done. I had to install a package to be able to recursively map through the children. Otherwise, the `CloseButton` couldn't be found

acd02: Using the `react-children-utilities` causes error in our CI, so instead I added the `deepFind` util as new **internal-utils**, see #1965

Powerplex (approved):

soykje (approved):

PR #2005, opened 2024-03-28
fix: Support inverted switch order

danieleloscozzese:

We use the Switch in Subito, with a UI where labels are to the left of the inputs as standard.
We noticed a problem on one of our pages where the page was horizontally scrollable whenever the switches were checked.

We traced this to the translation of the `<span>` which control the circular toggle: when checked, a wrapper moves outside (to the left) of the element itself and overflows.

This is easily reproducible when adding `flex-row-reverse` to the Switch as a class.

We saw no problems when adding the `overflow-x-hidden` as a general class, with no logic.

#### UI Before: Off

The UI before, which is all fine when the switch is not checked

#### UI Before: On

The UI before, which has horizontal overflow when the switch is checked

#### UI After: On

The UI after, which is all fine when the switch is checked

Co-authored-by: Chao Fang <lucafang@hotmail.it>
Co-authored-by: Tommaso Ruscica <ruscica@hotmail.it>


Powerplex:
Thanks for contributing. Don't worry about the failing CI steps it's because you merge from a fork and those steps can't work from a fork (for now).

I agree with the fix (`overflow-x-hidden`) but we are discussing if we should have stories to display a TW class (`flex-row-reverse`). 

It's not the first time developers needs this class to reverse the contents, but it happens a lot also with `Checkbox` and `Radio` too. Instead of a story, we are discussing adding an `inverse` prop that would apply that class internally. This would be documented in the props table at the top (with a description) instead of a dedicated story.

danieleloscozzese:
Understood, I'll drop the story here and update the PR.

Powerplex:
Thanks :) You can check the addition of the `reverse` prop on `switch/checkbox/radio` here: https://github.com/adevinta/spark/pull/2007

Powerplex (approved):

Issue #1957, opened 2024-03-12
[Button] Loading state - Update the specs to allow for the default option of having only the spinner on the left


Bibisirius:
We've decided to offer the option to display the `spinner` only on the left. This will maintain consistency with Android constraints.

![Image](https://github.com/adevinta/spark/assets/139967199/a27ea601-8b5a-460c-9279-db66b35fd287)

Powerplex:
We met between web devs and decided that there is no reason to hardcode this limitation in the code as this will add unnecessary complexity. It can remain a Spark/Polaris guidelines and no change is necessary.

Bibisirius:
Okay, noted @Powerplex ! 
In order to maintain consistency with the apps, we'll need to synchronize with the specs only for this point. No need to code it, just remove the loading right in the specs.

PR #2020, opened 2024-04-03
feat(button): remove the isLoading spinner placement fixing it as a trailing icon of the loadingText

andresin87:
removes the spinner placement of the button

BREAKING CHANGE: removes the spinner placement prop of the button not allowing to set it as a leading icon when loading state is abled
#1957

**TASK**: 

### Types of changes


packages/components/button/src/Button.stories.tsx
```diff
@@ -141,11 +141,7 @@ export const LoadingWithText: StoryFn = () => {
         Button with long text
       </Button>
 
-      <Button
-        isLoading={isLoading}
-        loadingText="Loading (spinner to the right)..."
-        spinnerPlacement="right"
-      >
+      <Button isLoading={isLoading} loadingText="Loading (spinner to the right)...">
```
Powerplex, line 144: you can remove the button entirely because the one above it does the same thing

acd02 (approved):

Powerplex (approved):

PR #2024, opened 2024-04-03
fix(dropdown): z-index conflict between dialog and dropdown

Powerplex:

We portalled `Dropdown.Popover` to fix a bug, but it introduced another bug: Dropdowns are now rendered portalled like `Dialog` so they share the same stacking context.

This is bad because their z-index are:
```
dropdown: '1000',
modal: '1400',
```

So now, a Dropdown inside a dialog is visually hidden behind it.

Instead of hard-coding the wrapping of `Dropdown.Popover` with a `Portal`, I added `Dropown.Portal` as part of the compound to make it optional. That way by default a dropdown is not portalled and inherits the closest stacking context.

A dropdown inside a dialog has a z-index of `1000` INSIDE a stacking context of `1400`.

### Types of changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2382,6 +2382,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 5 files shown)


acd02 (approved):

soykje (approved):
:fire_engine:

Issue #2049, opened 2024-04-16
Deprecate Radix useId for React version


soykje:
Radix implementation of `useId` React hook only seems to bring React <18 support, which we don't care. So we may want now to use the React one, instead.

PR #2052, opened 2024-04-16
refactor: deprecate radix useId for react one

soykje:
**TASK**: #2049 

To avoid duplicate ID issues with compound pattern, we want to prefix our generated IDs, and also to use React `useId` instead of Radix custom implementation, which doesn't seem to be relevant to our usage.

### Types of changes
- [x] 🧠 Refactor

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2381,6 +2381,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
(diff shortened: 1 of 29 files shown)


andresin87 (approved):
dnt know how lerna will deal with all that changes

acd02 (approved):

soykje:
> dnt know how lerna will deal with all that changes

Let's find out :nerd_face:

PR #2054, opened 2024-04-17
fix(combobox): address pointerEvents related bug

acd02:
This PR addresses a bug:
Which cannot be reproduced when running Storybook locally, in scenarios such as when a Dropdown is nested within a Dialog, the "props" object, containing styles computed by Radix may erroneously contain "pointerEvents = 'none'", while the Dropdown is open, making it impossible to select a value using a pointer device

### DIFF
--- a/packages/components/dropdown/src/DropdownItems.tsx
+++ b/packages/components/dropdown/src/DropdownItems.tsx
@@ -9,6 +9,17 @@ interface ItemsProps {
   className?: string
 }
 
+/**
+ * BUGFIX
+ *
+ * 1. The !pointer-events-auto class is needed to prevent a bug
+ *    which cannot be reproduced when running Storybook locally,
+ *    in scenarios such as when a Dropdown is nested within a Dialog,
+ *    the "props" object, containing styles computed by Radix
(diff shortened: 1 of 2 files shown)


packages/components/dropdown/src/DropdownItems.tsx
```diff
@@ -9,6 +9,17 @@ interface ItemsProps {
   className?: string
 }
 
+/**
+ * HACKS
```
soykje, line 13: I am not sure that we need to declare your solution as "hacky" :thinking: For me it is a simple bugfix, right?

soykje (approved):

packages/components/dropdown/src/DropdownItems.tsx
acd02: Yes, it's not exactly a hack, but I added this comment for two reasons. First, this bug cannot be reproduced locally, and secondly, we added this class to override styles computed not by us, but by an external library.

What would you suggest in place of HACK?

acd02: I went with BUGFIX

PR #2123, opened 2024-05-13
fix(combobox): placeholder only appears when there is no value on combobox

andresin87:
**TASK**: 

### Types of changes


packages/components/combobox/src/ComboboxInput.tsx
```diff
@@ -75,6 +75,10 @@ export const Input = forwardRef(
       ref: inputRef,
     })
 
+    const { selectedItems, selectedItem, multiple } = ctx
```
Powerplex, line 78: I favor if we don't destructure from `ctx`, especially if it's only used once. As the code get more complex, it helps see at first glance what comes from the context vs the props.

packages/components/combobox/src/ComboboxInput.tsx
```diff
@@ -75,6 +75,10 @@ export const Input = forwardRef(
...
+
+    const hasPlaceholder = multiple ? selectedItems.length === 0 : selectedItem === null
```
andresin87, line 80:
```suggestion
    const hasPlaceholder = ctx.multiple ? ctx.selectedItems.length === 0 : ctx.selectedItem === null
```

andresin87: @Powerplex ?

Powerplex (approved):

elisa-hery:
LGTM, @adevinta/spark-design do you want to check it ?

JeremiasUX:
Yes, looks good to me too. Thanks :)

Issue #2183, opened 2024-05-30
[Component] - Combobox - users not pleased with default filtering behaviour


Powerplex:
**### Current behaviour:** 

by default with autoFilter  + single selection, when you select a value in the menu, it becomes the value of the input and filters out all other values not matching. This is how most Combobox  are implemented (https://component.gallery/components/combobox/), but I agree :100: that it's not very practical.

**### Proposal:** 

Internally the component should be able to detect when the user started typing to start filtering. 

Once a value is selected, then the filtering behaviour is disabled/reset, so the user has directly access to the full list of values to select another one.

**### Next step:** 

- We must analyze all features (autoSuggest, autoComplete, search behaviour, etc. to check for regressions)
- Maybe we should not update the default behaviour but provide both behaviour ?

PR #2184, opened 2024-05-30
feat(combobox): improved UX of default combobox filtering behaviour

Powerplex:
**TASK**: #2183 

1. Read attached issue.
2. Proposal is to replace `autoFilter` (boolean), by `filtering: "auto" | "none" | "strict"`

`strict` matches the previous default behaviour, `auto` is the new/improved way for filtering, where the filtering only applies once the user starts typing.

For example, if the user selects a value in the list, it will reset/cancel the filtering so they can explore the full list of values.

### Types of changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2301,6 +2301,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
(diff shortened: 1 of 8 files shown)


acd02 (approved):

soykje (approved):
:fire:

Issue #2136, opened 2024-05-15
[Storybook] Check the read only as there is a bug


Bibisirius:
We currently have the option of having a disabled and enabled style by default in the Select.

Example: ![Image](https://github.com/adevinta/spark/assets/139967199/2a9802d3-6a40-46e6-9196-e4a65903897b)

There is currently a bug in the read-only style on the select, which needs to be corrected to maintain consistency.

- By default, the read-only style will be disabled. 
- If necessary, we can also use the enable version.

![Image](https://github.com/adevinta/spark/assets/139967199/9029be0e-ff43-4379-a612-a365a3bde08a)

soykje:
If I'm not wrong we also have similar issue with `disabled` state for the `Select`. Both will be fixed in this ticket 👍

soykje:
@Bibisirius cc @JeremiasUX https://spark-ui-uhip-k18p5l0o8-sparkteams-projects.vercel.app/?path=/docs/components-select--docs#read-only 👍

Bibisirius:
Check and approved ✅

PR #2198, opened 2024-06-04
fix(select): fix disabled and readonly styles

soykje:
**TASK**: #2136 

On the `Select` component `disabled` and `readOnly` states seem to be desync with input styles. We want to harmonize and fix this.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### Screenshots - Animations
image
image

### DIFF
--- a/packages/components/select/src/SelectTrigger.styles.tsx
+++ b/packages/components/select/src/SelectTrigger.styles.tsx
@@ -3,8 +3,8 @@ import { cva } from 'class-variance-authority'
 export const styles = cva(
   [
     'relative flex w-full items-center justify-between',
-    'min-h-sz-44 rounded-lg bg-surface text-on-surface px-lg',
-    'text-body-1',
+    'min-h-sz-44 rounded-lg px-lg',
+    'text-body-1 bg-surface text-on-surface',
     // outline styles
     'ring-1 outline-none ring-inset',
   ],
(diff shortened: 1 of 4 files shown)


packages/components/select/src/SelectTrigger.styles.tsx
```diff
@@ -17,11 +17,11 @@ export const styles = cva(
         success: 'ring-success',
       },
       disabled: {
-        true: 'disabled:bg-on-surface/dim-5 cursor-not-allowed text-on-surface/dim-3',
+        true: '!bg-on-surface/dim-5 cursor-not-allowed !text-on-surface/dim-3',
```
acd02, line 20: Do we really need to use the `!` variant here?

soykje: Yes, because of L7, where we define the `bg-surface` :disappointed:

soykje: I tried this first, but well then compound logic had to be also done for `disabled` variant... Seemed simpler this way.But if it is blocking to you I can totally try again :+1:

acd02: It's not a blocker on my side, but if we can avoid using the `!` 🤷

acd02 (approved):

Powerplex (approved):

andresin87 (approved):

Issue #2205, opened 2024-06-06
[Component] Checkbox - required attribute issue on checkboxes


acd02:
When a `CheckboxGroup` is marked as required, we should remove the `required` attribute from individual checkboxes as soon as one checkbox is checked. Currently, this is not the case.

PR #2206, opened 2024-06-06
fix(checkbox): remove the required attribute from individual checkboxes as soon as 1 checkbox is checked

acd02:
**TASK**:#2205

When a `CheckboxGroup` is marked as required, we should remove the required attribute from individual checkboxes as soon as one checkbox is checked.

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/checkbox/src/Checkbox.tsx
+++ b/packages/components/checkbox/src/Checkbox.tsx
@@ -2,7 +2,7 @@
 import { useFormFieldControl } from '@spark-ui/form-field'
 import { useMergeRefs } from '@spark-ui/use-merge-refs'
 import { cx } from 'class-variance-authority'
-import { forwardRef, useId, useRef } from 'react'
+import { forwardRef, useId, useMemo, useRef } from 'react'
 
 import { CheckboxGroupContextState, useCheckboxGroup } from './CheckboxGroupContext'
 import { CheckboxInput, CheckboxInputProps } from './CheckboxInput'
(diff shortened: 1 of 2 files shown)


andresin87 (approved):

packages/components/checkbox/src/Checkbox.tsx
```diff
@@ -79,6 +79,12 @@ export const Checkbox = forwardRef<HTMLButtonElement, CheckboxProps>(
       checkboxIntent: intentProp,
     })
 
+    const getIsRequired = () => {
```
soykje, line 82: Wdyt of using a memoized const instead? It might be more understandable code readers, and  would also take care of possible multiple toggles within a checkbox group :+1:

Powerplex (approved):

Powerplex:
Thanks. I checked and HTML `required` attribute is not supported. Instead everywhere it is suggested to use JS to manage it, so the fix is aligned with that.

packages/components/checkbox/src/Checkbox.tsx
acd02: I'm not sure I understand, you mean something like this?:
```tsx
const isRequired = useMemo(() => {
  if (!group) return isRequired

  return isRequired ? !group.value?.length : false
}, [group, isRequired])
```

soykje: Yes!

Issue #2208, opened 2024-06-07
Combobox Type for ComboboxPropsSingle shouldn't be undefined


christophesube:
### Description

When I pass a value to onValueChange Combobox Props, this one should not be able to be undefined.

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Spark Component and Version

@spark-ui/combobox: 1.1.2

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

PR #2209, opened 2024-06-07
fix(combobox): removed unnecessary undefined type for combobox onValu…

Powerplex:
…eChange

**TASK**: #2208 

`onValueChange` accepted `string | undefined` as a first parameter in its callback (single selection mode only).

It means that our consumer are forced to check the type:
```
<Combobox value={value} onValueChange={(value)  => {
  if (value !== undefined) setValue(value)
}}
```

But I think such a case can't happen. Once a Combobox is controlled, its `value` must remain a `string`.

### Types of changes

### DIFF
--- a/packages/components/combobox/src/Combobox.stories.tsx
+++ b/packages/components/combobox/src/Combobox.stories.tsx
@@ -922,16 +922,12 @@ export const ModalSearch: StoryFn = () => {
           <Dialog.Content size="sm">
             <Combobox
               onValueChange={value => {
-                if (value === undefined) {
-                  setValue(value)
-                } else {
-                  const [_, id] = value.split('-')
-                  const book = books.find(({ id: bookId }) => `${bookId}` === id)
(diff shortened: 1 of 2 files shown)


packages/components/combobox/src/Combobox.stories.tsx
```diff
@@ -922,16 +922,12 @@ export const ModalSearch: StoryFn = () => {
...
-
-                  if (book) {
-                    setValue(book?.name)
-                    setIsOpen(false)
-                  }
+                const [_, id] = value.split('-')
+                const book = books.find(({ id: bookId }) => `${bookId}` === id)
+
+                if (book) {
+                  setValue(book?.name)
```
acd02, line 929:
```suggestion
                  setValue(book.name)
```

acd02 (approved):

soykje (approved):

Issue #2199, opened 2024-06-04
[Component] - Snackbar


acd02:
If you watch the video below, you will understand the bug. Essentially, if you slightly swipe the snackbar without swiping enough to close it, the `onClose` button fails to close the snackbar.

Additionally, the `timeout` function that should automatically close the snackbar after 5 seconds also becomes inactive.
When in this state, the only way to close the snackbar is to swipe it

![Image](https://github.com/adevinta/spark/assets/51230231/0f7479fc-e103-45f6-b49d-0ec70c256aa2)

soykje:
@elisa-hery Ready for QA time: https://spark-ui-uhip-n4q4w74h4-sparkteams-projects.vercel.app/iframe.html?id=components-snackbar--close&viewMode=story 🙏

elisa-hery:
@soykje I tested all [these cases](https://docs.google.com/spreadsheets/d/1plmBTjbBxT7SyDgrxlRjYHvlS9fYjJVRlGwbpet-HoM/edit?gid=1435862231#gid=1435862231) on safari web / safari mobile / chrome web : it's all good, this bug is fixed ✅

Nevertheless it made me notice 2 things : 

1. on **safari** (Version 17.4.1), with keyboard navigation, enter or space don't trigger action or close button (I know you cannot work on safari, but at least maybe I should show you)
2. with a long text in the action button, the button takes the whole space (and also it's not vertically aligned), and I'm not sure this is what we want : 
<img width="467" alt="Screenshot 2024-06-18 at 11 30 45" src="https://github.com/adevinta/spark/assets/142396647/046d2fbc-ec53-44a1-b0b4-2368b262b835">

maybe we can talk about this !

elisa-hery:
1. we created a new task for this --> https://github.com/adevinta/spark/issues/2249 
2. @soykje fixed it ✅

PR #2223, opened 2024-06-11
fix(snackbar): fixes and improves snackbar swipe styles to avoid translate collision

soykje:
**TASK**: #2199 

Snackbar animations (with `translate` property) were having edge case issue with collission between different data attributes bringing their own translate rule.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### DIFF
--- a/packages/components/snackbar/src/Snackbar.tsx
+++ b/packages/components/snackbar/src/Snackbar.tsx
@@ -55,7 +55,13 @@ export const Snackbar = forwardRef<HTMLDivElement, SnackbarProps>(
     useEffect(() => {
       addProvider(ref)
 
-      return () => deleteProvider(ref)
+      return () => {
+        for (const toast of getGlobalSnackBarQueue().visibleToasts) {
+          toast.animation = undefined
+        }
+
+        deleteProvider(ref)
+      }
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [])
 
--- a/packages/components/snackbar/src/SnackbarItem.tsx
+++ b/packages/components/snackbar/src/SnackbarItem.tsx
@@ -95,7 +95,6 @@ export const SnackbarItem = forwardRef<HTMLDivElement, PropsWithChildren<Snackba
       onSwipeCancel: state.resumeAll,
       onSwipeEnd: ({ direction }) => {
         ;['left', 'right'].includes(`${direction}`) && state.close(toast.key)
-        state.resumeAll()
       },
     })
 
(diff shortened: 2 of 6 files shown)


andresin87 (approved):

acd02 (approved):

packages/components/snackbar/src/Snackbar.tsx
```diff
@@ -55,7 +55,13 @@ export const Snackbar = forwardRef<HTMLDivElement, SnackbarProps>(
     useEffect(() => {
       addProvider(ref)
 
-      return () => deleteProvider(ref)
+      return () => {
+        for (const toast of getGlobalSnackBarQueue().visibleToasts) {
+          toast.animation = undefined
```
soykje, line 60: More sugar than real improvement... This allows us to ensure `toast.animation` state is properly resetted on unmount.

packages/components/snackbar/src/SnackbarItem.tsx
```diff
@@ -131,6 +134,19 @@ export const SnackbarItem = forwardRef<HTMLDivElement, PropsWithChildren<Snackba
...
+      if (!ref.current) return
+
+      /**
+       * After canceling swipe gesture, we don't want to keep related
+       * data-attributes anymore (to preserve animations & transitions)
+       */
+      if (swipeState === 'cancel' && toast.animation === 'exiting') {
+        ref.current.removeAttribute('data-swipe')
+        ref.current.removeAttribute('data-swipe-direction')
+      }
```
acd02, line 147: Couldn't we use the `onSwipeCancel` callback here, instead of this `useEffect`?

soykje: Sadly not :disappointed: If I'm not wrong, we don't "listen" to `toast.animation` changes using the `onSwipeCancel` callback.
I tried using `useCallback`, ... with no success. But as said in Slack I'll be glad to get your eyes on this :+1:

acd02: I checked out your branch, and I don't think you need a `useEffect`, instead you could handle that directly in your render:
```tsx
data-swipe={{ ...(toast.animation === 'exiting' && swipeState) }}
data-swipe-direction={{ ...(toast.animation === 'exiting' && swipeDirection) }}
```

soykje: After testing it seems that your suggestion doesn't work :sweat_smile: I guess this is because we need to wait the browser to "paint the screen" (CSS) before removing these attributes.

soykje: Ok copy-pastes were too lazy, but it finally works :+1: Thx @acd02

acd02 (approved):

Issue #2224, opened 2024-06-11
[Component] - Collapsible animation


Powerplex:
https://www.figma.com/design/0QchRdipAVuvVoDfTjLrgQ/Component-Specs-of-Spark?node-id=48556-33536&t=Zig26rHfRwlyX3AM-1

PR #2225, opened 2024-06-11
style(collapsible): collapsible animation

Powerplex:
**TASK**: #2224 

See attached ticket. Adding `Collapsible` animation using both opacity and height animation.

### Types of changes
- [x] 🧾 Documentation
- [x] 💄 Styles

### DIFF
--- a/packages/utils/tailwind-plugins/animations/index.js
+++ b/packages/utils/tailwind-plugins/animations/index.js
@@ -121,13 +121,13 @@ module.exports = plugin.withOptions(
     theme: {
       extend: {
         keyframes: {
-          collapseIn: {
-            '0%': { height: 0 },
-            '100%': { height: 'var(--height)' },
+          standaloneCollapseIn: {
+            '0%': { opacity: 0, height: 0 },
+            '100%': { opacity: 1, height: 'var(--height)' },
           },
-          collapseOut: {
-            '0%': { height: 'var(--height)' },
-            '100%': { height: 0 },
+          standaloneCollapseOut: {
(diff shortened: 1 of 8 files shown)


andresin87 (approved):
🚀

packages/utils/tailwind-plugins/animations/index.js
```diff
@@ -121,6 +121,14 @@ module.exports = plugin.withOptions(
     theme: {
       extend: {
         keyframes: {
+          collapseIn: {
```
acd02, line 124:
```suggestion
          standaloneCollapseIn: {
```

see 👉 https://github.com/adevinta/spark/wiki/Conventions-for-Adding-Custom-Styles-in-Tailwind#standalone-keyframes

acd02 (approved):

packages/utils/tailwind-plugins/animations/index.js
Powerplex: As it will also be re-used by `Accordion` soon, should it still be prefixed ?

acd02: Maybe we could name it `standaloneDisclosureIn` / `standaloneDisclosureOut`. Then, we could use that for both the collapsible and accordion elements, wdyt?

acd02: beware, you used  `standaloneCollapseIn` / `standaloneCollapseOut`, and not `standaloneDisclosureIn` / `standaloneDisclosureOut`

PR #2242, opened 2024-06-17
Scan adoption script 1

andresin87:
## utils(cli): Spark adoption scrip RC

**TASK**: 

### Types of changes

Add extra arguments to the script command

```js
program
  .command('adoption')
  .option('-c, --configuration <config>', 'configuration file route', '.spark-ui.cjs')
  .option('-o, --output <output>', 'output file route')
  .option('-v, --verbose', 'output log information', false)
  .option('-d, --details', 'output information about each match', config.details)
  .option('-s, --sort <sort>', 'sort results (count or alphabetical)', config.sort)
  .option('-dir, --directory <directory>', 'directory to parse', config.directory)
  .option('-ext, --extensions <extensions...>', 'file extensions to parse', config.extensions)
  .option('-i, --imports <imports...>', 'import patterns to identify', config.imports)
```
- [x] 🛠️ Tool


packages/utils/cli/src/scan/utils/logger.mjs
```diff
@@ -1,19 +1,50 @@
 import chalk from 'chalk'
 
-export const logger = {
+export class Logger {
```
soykje, line 3: Why not using `packages/utils/cli/src/core/Logger.mjs` directly? It seems we duplicate things :thinking:

soykje (approved):

packages/utils/cli/src/scan/loadConfig.mjs
```diff
@@ -0,0 +1,32 @@
...
+    } else {
+      logger.warning('⚠️ No custom configuration file found')
+      logger.info('ℹ️ Loading default configuration')
+
+      return { ...defaultConfig }
+    }
+  } catch (error) {
+    logger.error('💥 Something went wrong loading the custom configuration file')
+
+    return { ...defaultConfig }
```
soykje, line 30: Don't you think it would be clearer for CLI user to throw an error and exit here?
For now we have no details on what was the failure, and as we fallback to default config, user may not notice his config is wrong :thinking:

packages/utils/cli/src/scan/loadConfig.mjs
```diff
@@ -0,0 +1,32 @@
...
+
+export async function loadConfig(configFileRoute, { logger }) {
+  try {
+    if (existsSync(configFileRoute)) {
+      logger.info('ℹ️ Loading spark-ui custom configuration file')
+      const { default: customConfig } = await import(configFileRoute)
+
+      const config = {
+        adoption: merge(defaultConfig, {
+          ...customConfig.adoption,
```
soykje, line 14: After testing, this doesn't seem to work, with a custom file using the same structure than `./config.mjs` :confused: Happy to discuss :+1:

soykje (changes requested):

andresin87 (commented):
✅

packages/utils/cli/src/scan/loadConfig.mjs
soykje: @andresin87 If I'm not wrong, we still return a config on `loadConfig` failure? As said, don't you think we should rather throw error and exit the process, without returning a value?
Actually if would be difficult for the user to identify that he/she defined an invalid configuration file, if the process still goes on. He/she would end with results but not expected ones :/

soykje: @andresin87 We still don't kill the process here, and return a config, which still could be confusing... That said, I'm not sure we really need this `try...catch` as `path.join(process.cwd(), configuration || '.spark-ui.cjs')` shouldn't throw any error, wdyt? cc @acd02 (other eyes may be interesting here)

acd02: I must admit that just by looking at the code like this, I don’t think I’ll have anything relevant to add 😅 

I would have to checkout your branch and throw some `console.log` around to get a better gist at what exactly this code is doing

soykje (approved):

PR #2262, opened 2024-06-21
style(accordion): disable collapse icon transition when prefers motio…

Powerplex:

Disable smooth rotate animation when `prefers-reduced-motion: reduce` is set in the user preferences.

### Types of changes
- [x] 💄 Styles

### DIFF
--- a/packages/components/accordion/src/AccordionItemTrigger.tsx
+++ b/packages/components/accordion/src/AccordionItemTrigger.tsx
@@ -43,7 +43,7 @@ export const ItemTrigger = forwardRef<HTMLButtonElement, AccordionItemTriggerPro
         <div className="flex items-center gap-lg">{children}</div>
         <Icon
           intent="neutral"
-          className={cx('shrink-0 rotate-0 transition duration-100 ease-in', {
+          className={cx('shrink-0 rotate-0 duration-100 ease-in motion-reduce:transition-none', {
             'rotate-180': isOpen,
           })}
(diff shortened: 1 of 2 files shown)


soykje (approved):
I'm wondering if we shouldn't handle this more "globally", for all our components (via a custom class maybe?)

packages/components/accordion/src/AccordionItemTrigger.tsx
```diff
@@ -43,9 +43,12 @@ export const ItemTrigger = forwardRef<HTMLButtonElement, AccordionItemTriggerPro
         <div className="flex items-center gap-lg">{children}</div>
         <Icon
           intent="neutral"
-          className={cx('shrink-0 rotate-0 transition duration-100 ease-in', {
-            'rotate-180': isOpen,
-          })}
+          className={cx(
+            'shrink-0 rotate-0 transition duration-100 ease-in motion-reduce:!transition-none',
```
acd02, line 47: Why use the `!` modifier here?

You can write like this instead:
```tsx
shrink-0 rotate-0 motion-reduce:transition duration-100 ease-in
```

Powerplex: Wasn't aware of this `motion-safe`, I will do that and try.

acd02 (approved):

andresin87 (approved):

Issue #2243, opened 2024-06-17
[Component] Tag: New intent "Surface"


JeremiasUX:
We received a request via slack:

- https://adevinta.slack.com/archives/C03GNJ2A4HW/p1715588690933379

Here is the jira ticket with more information:

- https://jira.ets.mpi-internal.com/browse/DO-1740

Spec:
- https://www.figma.com/design/0QchRdipAVuvVoDfTjLrgQ/Component-Specs-of-Spark?node-id=1613-33323&t=bcMEOdvB4JqTON6Z-4

Goal:
- Provide a new intent - surface

PR #2279, opened 2024-06-28
feat(tag): add "surface" intent

acd02:
**TASK**: #2243

Add surface intent according to updated design specs

### Types of changes
- [ ] 🛠️ Tool
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/packages/components/tag/src/Tag.stories.tsx
+++ b/packages/components/tag/src/Tag.stories.tsx
@@ -24,6 +24,7 @@ const intents: TagProps['intent'][] = [
   'danger',
   'info',
   'neutral',
+  'surface',
 ]
 const designs: TagProps['design'][] = ['filled', 'outlined', 'tinted']
 
@@ -43,11 +44,17 @@ export const Intent: StoryFn = _args => (
   <div className="flex flex-col gap-md">
     {designs.map(design => (
       <div key={design} className="flex flex-row gap-md">
-        {intents.map(intent => (
-          <Tag key={intent} design={design} intent={intent}>
-            {intent} tag
-          </Tag>
-        ))}
+        {intents.map(intent => {
+          if (design !== 'filled' && intent === 'surface') {
+            return <span className="self-center text-small">N/A</span>
+          }
+
+          return (
+            <Tag key={intent} design={design} intent={intent as any}>
+              {intent} tag
+            </Tag>
+          )
+        })}
       </div>
     ))}
   </div>
--- a/packages/components/tag/src/Tag.tsx
+++ b/packages/components/tag/src/Tag.tsx
@@ -3,7 +3,7 @@ import { forwardRef, type PropsWithChildren } from 'react'
 
 import { tagStyles, type TagStylesProps } from './Tag.styles'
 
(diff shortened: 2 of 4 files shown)


packages/components/tag/src/Tag.stories.tsx
```diff
@@ -43,11 +44,17 @@ export const Intent: StoryFn = _args => (
...
-            {intent} tag
-          </Tag>
-        ))}
+        {intents.map(intent => {
+          if (design !== 'filled' && intent === 'surface') {
+            return <span className="self-center text-small">N/A</span>
+          }
+
+          return (
+            <Tag key={intent} design={design} intent={intent as any}>
```
acd02, line 53: Sorry about the `any` 😅, but without it, TS will complain because it can't infer that due to my previous `if` block, we cannot have a surface of **intent**, with a design of **outlined** or **tinted** when we end up here

acd02:
if you wanna check @JeremiasUX ⬇️  ,
https://spark-ui-uhip-kerj6x6jz-sparkteams-projects.vercel.app/?path=/docs/components-tag--docs

packages/components/tag/src/Tag.tsx
```diff
@@ -12,6 +12,21 @@ export interface TagProps
...
+  K extends TagStylesProps['intent'] | never = never,
+> {
+  design?: Design
+  intent?: Exclude<TagStylesProps['intent'], K>
+}
+
+type ValidTagDesignIntent =
+  | FilteredDesignIntent<'tinted', 'surface'>
+  | FilteredDesignIntent<'outlined', 'surface'>
+  | FilteredDesignIntent<'filled'>
```
acd02, line 26: What is being done here is to prevent unwanted combinations.
The surface **intent** can only be applied if the design is set to **filled**.
We cannot have the following:
```tsx
<Tag design="outlined" intent="surface">
```

packages/components/tag/src/variants/filled.ts
```diff
@@ -46,4 +46,9 @@ export const filledVariants = [
     design: 'filled',
     class: tw(['bg-neutral', 'text-on-neutral']),
   },
+  {
+    intent: 'surface',
+    design: 'filled',
+    class: tw(['bg-surface', 'text-on-surface']),
```
soykje, line 52: Do we still need this `tw()` util?

packages/components/tag/src/Tag.stories.tsx
soykje: You should be ashamed!

soykje (approved):

packages/components/tag/src/variants/filled.ts
acd02: Unfortunately, yes. Otherwise, we lose the autocompletion / intellisense. This is due to a limitation of the Tailwind ESLint plugin's typings, which cannot handle very large objects, if I recall correctly

andresin87 (approved):

PR #2312, opened 2024-07-19
[chore] deprecate icons first pass

soykje:
As Spark icons are not provided directly anymore, we want to clean our directory. This is a first pass, and the remaining icons may be removed pairing with rest of the team for a global agreement. Here e only focus on : non compliant icons (stroke too thin regarding to original icon set), too strongly product-related icons and duplicate ones :+1: 

### Types of changes
- [x] 🧠 Refactor

### Screenshots - Animations
image


soykje:
For review/testing purpose I made available a "ghost" link: `<host>/iframe.html?id=components-icons--list&viewMode=story`

andresin87 (approved):

Powerplex:
My only concern (and maybe we should do a pair review on Storybook directly) is regarding the icons that had the correct "style". In my opinion some of them deserved to be deleted because:
- duplicates (or too close to another icon)
- too product-specific.

But some of them could maybe be renamed instead of deleted. And some of them are not that "product specific":
```
Bank
Calculate
Car
CarWarranty
Country
DeliveryFast
DeliveryHands
DeliveryTruck
Dissatisfied
facebook
Flashlight
Holiday (for example, this is an umbrella icon, could be renamed to "Umbrella")
Hotel
Money
Neutral
PiggyBank
ProCursor
Pro
Sad
Satisfied
Security
SecurityProfileAccount
SecurityProfile
ShoppingCart
Speedmeter
Store
Support
Train
Walker
WarningSecurity
```

All those (their fill/outlined versions) we could review during a live session just to be sure.

PR #2324, opened 2024-07-26
test: setup a11y tests with playwright

soykje:
**TASK**: #2327 

Following Playwright setup, we want here to try an update on ou a11y testing process, by using Playwright+Axe instead of a custom script using Axe.
Our goal:

- a maybe (?) **faster and robuster solution** relying on Playwright solution => This will be testing during some weeks, comparing CI results from both solutions (current and PW one)
-  a cleaner testing solution, with **specific tests/implementations for a11y testing**. Storybook stories would remain... Storybook stories. This would help us target with more accuracy our specific needs to be confident in our a11y quality :+1: 

#### What is proposed here:
- [x] a dedicated process to test component's implementations pages in a fluid (?) process
- [x] two specific CLI commands for E2E and A11Y
- [x] a refacto of playwright CI actions and implementation within our quality workflow, to target specifically E2E and A11Y
- [x] ...

### Types of changes
- [x] 🛠️ Tool
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor


soykje:
:information_source: For now we will only check for violations here, as there seems to be an issue with Axe tooling. Despite of this the setup stays relevant and the fix should come soon :+1:

.github/composite-actions/pw-a11y/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  using: composite
+  steps:
+    - name: Install
+      uses: ./.github/composite-actions/install
+
+    - name: Install Playwright Browsers
+      shell: bash
+      run: npx playwright install --with-deps
+
+    - name: Run tests
```
acd02, line 13:
```suggestion
    - name: Run a11y tests
```

.github/workflows/quality.yml
```diff
@@ -117,6 +117,28 @@ jobs:
         if: steps.files.outputs.all_changed_files_count > 0
         uses: ./.github/composite-actions/a11y
 
+  a11y-with-playwright:
```
acd02, line 120: Does this mean this will run on every PR? If that’s the case, I’m not sure it’s called for, since they’re accessibility tests rather than functional tests.

Maybe we could run them like cron jobs, twice a week or something like that, wdyt?

.github/composite-actions/pw-a11y/action.yml
soykje: This name would appear at a very nested depth. Not sure it is so important, but nothing blocking for me.

.github/workflows/quality.yml
soykje: Nope, that is the whole point of this: https://github.com/adevinta/spark/pull/2324/files#diff-b31f1e5654d027d84c24cdba2ff31ab79e3ea617091ed348a97476a27b076078R139 :+1: 
Playwright tests (just as initial a11y one) would only be triggered on specific file changes. Else it would skip.

acd02 (approved):

packages/components/dialog/src/Dialog.stories.tsx
```diff
@@ -50,6 +50,42 @@ export const Default: StoryFn = () => {
   )
 }
 
+export const QAWithDefaultOpen: StoryFn = () => {
```
Powerplex, line 53: I think we should remove this story (and all QA related stories)

Powerplex (approved):

PR #2330, opened 2024-07-29
refactor: split tests into multiple steps

acd02:
https://x.com/housecor/status/1816867564796117228

### Types of changes
- [x] 🛠️ Tool
- [x] 🧪 Test
- [x] 🧠 Refactor


andresin87 (approved):

soykje (approved):

Powerplex:
I am confused (even after reading the tweet) about why it improves performance, as you were already using `page.goto()` only once per test.

How does `test.step` improves things ? Reading the doc I think it only improves readability by splitting long tests into smaller tests. Am I right that our tests were already performant because page was already loaded only once ?

PR #2350, opened 2024-08-06
test: adds more implementations for pw+axe testing

soykje:
**TASK**: #2351 

Adds more implementations to test accessibility with PW+Axe workflow :muscle: 

### Types of changes
- [x] 🧪 Test

### DIFF
--- a/e2e/a11y/pages/FormField.tsx
+++ b/e2e/a11y/pages/FormField.tsx
@@ -1,0 +1,48 @@
+import { FormField } from '@spark-ui/form-field'
+import React from 'react'
+
+export const A11yFormField = () => (
+  <section>
+    <div>
+      <FormField name="email">
+        <FormField.Label>Email</FormField.Label>
+
+        <FormField.Control>
+          {({ id, name, description }) => (
+            <input
+              type="email"
+              id={id}
+              name={name}
+              aria-describedby={description}
+              className="rounded-sm border-md border-neutral p-md outline-none"
+            />
+          )}
+        </FormField.Control>
+
+        <FormField.HelperMessage>We will never share your email</FormField.HelperMessage>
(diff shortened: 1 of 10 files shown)


e2e/a11y/pages/FormField.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const A11yFormField = () => (
+  <section>
+    <div>
+      <FormField name="email">
+        <FormField.Label>Email</FormField.Label>
+
+        <FormField.Control>
+          {({ id, name, description }) => (
+            <input
```
acd02, line 12: you should use a `@spark-ui/input` here

e2e/a11y/pages/FormField.tsx
```diff
@@ -0,0 +1,48 @@
...
+      </FormField>
+    </div>
+
+    <div>
+      <FormField name="password" isRequired>
+        <FormField.Label>Password</FormField.Label>
+
+        <FormField.Control>
+          {({ id, name, description, isRequired }) => (
+            <input
```
acd02, line 32: same here

e2e/a11y/pages/FormField.tsx
soykje: Why? This test should focus on form-field, shouldn't it?

acd02: Yes, but why not make sure the `Formfield` works well with our input at the same time?

soykje: Tbh I'm not sure... In this case shouldn't we add all our input components within Formfield implementation? :thinking:

acd02 (approved):

Powerplex (approved):

PR #2382, opened 2024-08-26
fix: use moduleResolution 'bundler' instead legacy 'node' (node10) for correct support of ESM

yspychala:
#### Context

We recently migrated our project to ESM. However, we still use `node` (= node10 and older) as the module resolution strategy for TypeScript, which is meant for CommonJS `require`.
Since we use modern Node.js (v22.3.0), we can switch to `bundler`.

#### Issue
If we actually configure this in the project, various TypeScript errors appear and most of them are related to Spark and its use of `class-variance-authority`. It seems to concerns only packages using variants.

For example, there’s an error regarding `@spark-ui/radio-group`:

image

If we follow the error, we end up in `node_modules/@spark-ui/radio-group/dist/RadioInput.styles.d.ts` where the variants are defined. Then, if we hover the `RadioInputVariantsProps`, we can see some `any` that shouldn’t be there, which is the reason of all the errors.

image

We can notice that a type is imported from `class-variance-authority`:
```ts
import("class-variance-authority/dist/types").ClassProp
```

However, this path isn’t correctly resolved with the new module resolution strategy. If we look at the module’s package.json, we can see the following configuration:

```json
  "exports": {
    ".": {
      "import": "./dist/index.mjs",
      "require": "./dist/index.js",
      "types": "./dist/index.d.ts"
    },
    "./types": {
      "types": "./dist/types.d.ts"
    }
  },
  "main": "dist/index.js",
  "module": "dist/index.mjs",
  "types": "./dist/index.d.ts",
```

The path `/dist/types` doesn’t resolve to anything if we refer to the `exports` field.

#### Solution

The import should use `/types` instead of `/dist/types` to resolve the `types.d.ts` file. Picking correctly the file means we have to change the way the Spark packages are bundled.

If we look at the current TypeScript configuration, we can see `node` is used as a resolution strategy (remember it’s for node 10 and older versions). We could use `nodenext`, but since Vite is used here to emit the files and not TypeScript, we can safely use `bundler` instead.

By doing this, we can see that if we build again `@spark-ui/radio-group`, we now have the correct import as well as the correct type for the variants, and all the TS errors related to Spark vanish:

image

image

image

#### Additional changes
`rollup-plugin-terser` was used in Spark. However, this package is deprecated and not maintained anymore. `@rollup/plugin-terser` has been installed to replace this.
This change was needed because the legacy package wasn’t exporting correctly the types from an ESM point of view, thus making the typecheck fail where the package was used.

### Types of changes
- [x] 🛠️ Tool


yspychala:
Deployment job is failing because of missing secrets. Might be because of this pull request comes from a fork repository and the fork doesn’t have any rights.

PR #2395, opened 2024-09-05
feat(stepper): setting up stepper component with basic behavior

soykje:
**TASK**: https://jira.ets.mpi-internal.com/browse/SPA-48

Initialize component with basic behavior.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 💄 Styles

image

### DIFF
--- a/packages/components/stepper/src/Stepper.doc.mdx
+++ b/packages/components/stepper/src/Stepper.doc.mdx
@@ -1,5 +1,5 @@
 import { Meta, Canvas } from '@storybook/addon-docs'
-import { ArgTypes } from '@storybook/blocks';
+import { ArgTypes } from '@storybook/blocks'
 
 import { Stepper } from '.'
 
@@ -20,12 +20,15 @@ npm install @spark-ui/stepper
 ## Import
 
 ```tsx
-import { Stepper } from "@spark-ui/stepper"
+import { Stepper } from '@spark-ui/stepper'
 ```
 
 ## Props
 
 <ArgTypes of={Stepper} />
 
-## Variants
+## Usage
+
+### Default
+
 <Canvas of={stories.Default} />
--- a/packages/components/stepper/src/Stepper.doc.mdx
+++ b/packages/components/stepper/src/Stepper.doc.mdx
@@ -28,6 +28,19 @@ import { Stepper } from '@spark-ui/stepper'
 <ArgTypes
   of={Stepper}
   description="A container for all the stepper component parts."
+  exclude={[
+    'onPress',
+    'onPressStart',
+    'onPressEnd',
+    'onPressChange',
+    'onPressUp',
+    'onFocusChange',
+    'aria-labelledby',
+    'aria-describedby',
+    'aria-details',
+    'onKeyDown',
+    'onKeyUp',
+  ]}
   subcomponents={{
     'Stepper.Input': {
       of: Stepper.Input,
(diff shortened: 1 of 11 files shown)


packages/components/stepper/src/Stepper.doc.mdx
```diff
@@ -20,12 +21,113 @@ npm install @spark-ui/stepper
...
+import { Stepper } from '@spark-ui/stepper'
 ```
 
 ## Props
 
-<ArgTypes of={Stepper} />
+<ArgTypes
+  of={Stepper}
+  description="A container for all the stepper component parts."
+  exclude={[
```
Powerplex, line 32: I'm curious why we need to manually exclude this, are they not excluded by our global storybook filter by default ? (We didn't have to do this on other components)

packages/components/stepper/src/useStepper.ts
```diff
@@ -0,0 +1,37 @@
...
+}: UseStepperArgs): UseStepperReturn => {
+  const state = useNumberFieldState({
+    ...rest,
+    isDisabled: rest.disabled,
+    isReadOnly: rest.readOnly,
+    isRequired: rest.required,
+    locale,
+  })
+
+  const { groupProps, inputProps, incrementButtonProps, decrementButtonProps } = useNumberField(
```
Powerplex, line 19: Personal preference, but on other components we expose the whole api returned by the library (wether it is Zag, PopperJS, etc), it avoids destructuring, simplifies typing (eliminated the need to use `Pick<>` and allows us to use/explore the full API in each part of the compound.

WDYT ?

packages/components/stepper/src/Stepper.tsx
```diff
@@ -1,7 +1,144 @@
...
+      decrementAriaLabel,
+      ...stepperProps
+    },
+    forwardedRef
+  ) => {
+    const inputRef = useRef(null)
+
+    const {
+      groupProps,
+      inputProps: _inputProps,
```
Powerplex, line 40: Why the `_` prefix ? Why not also on `groupProps` ?

packages/components/stepper/src/Stepper.tsx
```diff
@@ -1,7 +1,144 @@
...
+        return childrenArray
+          .filter(isValidElement)
+          .find(child =>
+            (child.type as FC & { displayName?: string }).displayName?.includes(elementDisplayName)
+          )
+      },
+      [children]
+    )
+
+    const incrementBtnFromChildren = findElement('Stepper.IncrementButton')
```
Powerplex, line 66: The `findElement` pattern is a anti-pattern that goes against our composability rules, we decided to stop using it. Is there a reason you must use it ?

packages/components/stepper/src/useStepper.ts
soykje: Passing a huge and obscure pack of props could also be less readable I think. I guess it is a matter of personal preference but using (and exposing) only the necesary parts also allows us to keep our code more easy to read and understand.

Powerplex (approved):

packages/components/stepper/src/Stepper.doc.mdx
Powerplex: I agree with you, but I thought native react props were automatically omitted with our setup

packages/components/stepper/src/Stepper.tsx
Powerplex: Let's discuss it when you come back :) I agree it is necessary in some cases (I think only Select and Snackbar it is justified). I think it should be eliminated from Input too, but it is a breaking change and is not an emergency.

PR #2401, opened 2024-09-09
fix: compound export pattern messing with refs

soykje:
We noticed (!) that our export pattern for compound components was messing with ref prop on base component. We naturally want here to fix that error.

### Types of changes


packages/components/breadcrumb/src/Breadcrumb.stories.tsx
```diff
@@ -11,27 +12,31 @@ const meta: Meta<typeof Breadcrumb> = {
 
 export default meta
 
-export const Default: StoryFn = _args => (
-  <Breadcrumb aria-label="Breadcrumb">
-    <Breadcrumb.Item>
-      <Breadcrumb.Link href="/">Home</Breadcrumb.Link>
-    </Breadcrumb.Item>
+export const Default: StoryFn = _args => {
+  const ref = useRef(null)
```
acd02, line 16: is this `ref` useful?

acd02: Then maybe write a dedicated test?

Powerplex (approved):

packages/components/breadcrumb/src/Breadcrumb.stories.tsx
soykje: You made a point :+1:

PR #2418, opened 2024-09-18
style(form-field): inline required indicator placement

Powerplex:
**TASK**: [SPA-6](https://jira.ets.mpi-internal.com/browse/SPA-6)

- Update `FormField.Label` to not be a `flex` container, in order to inline the `RequiredIndicator` with the label text. 

### Types of changes
- [x] 💄 Styles

### Screenshots - Animations
Capture d’écran 2024-09-18 à 14 37 32

### DIFF
--- a/packages/components/form-field/src/FormFieldLabel.tsx
+++ b/packages/components/form-field/src/FormFieldLabel.tsx
@@ -38,7 +38,6 @@ export const FormFieldLabel = forwardRef<HTMLLabelElement, FormFieldLabelProps>(
         htmlFor={htmlFor}
         className={cx(
           className,
-          'flex items-center gap-sm',
           disabled ? 'pointer-events-none text-on-surface/dim-3' : undefined
         )}
         asChild={asChild}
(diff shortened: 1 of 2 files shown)


packages/components/form-field/src/FormFieldLabel.tsx
```diff
@@ -38,7 +38,6 @@ export const FormFieldLabel = forwardRef<HTMLLabelElement, FormFieldLabelProps>(
         htmlFor={htmlFor}
         className={cx(
           className,
-          'flex items-center gap-sm',
```
Powerplex, line 41: @acd02 I'm afraid removing this could visually break custom labels on Polaris side, wdyt ?

(But to do this ticket we should remove `flex` behaviour from the label.

acd02: There is only one occurrence in the Polaris codebase where a **flex-related** class is passed to the `FormField.Label`, so I think we're safe.

- src/components/InputPhone/wrapper.tsx

acd02 (approved):

Issue #2439, opened 2024-10-01
[Dialog] Border radius inconsistency when content overflows


acd02:
When using the `Dialog` component under the following conditions, the border radius is not applied correctly:

1. The `Dialog` has more content than its height can accommodate, causing a scrollbar to appear.
2. Without `Dialog.Header`: The top-right corner of the Dialog loses its border radius.
3. Without `Dialog.Footer`: The bottom-right corner of the Dialog loses its border radius.

**2:**
![Image](https://github.com/user-attachments/assets/a422d26a-4f90-446f-a6f9-a425135032c3)

**3:**
![Image](https://github.com/user-attachments/assets/c3118b05-371d-43d2-a7fe-f76a1fa90d03)

This PR aims to address the issue by ensuring the correct border radius is applied even when `Dialog.Header` or `Dialog.Footer` are not used.

**after:**
![Image](https://github.com/user-attachments/assets/b0198ae9-8fdf-4886-accf-d723ee711429)

PR #2441, opened 2024-10-01
feat(dialog): ensure correct border radius for overflowing content

acd02:
**TASK**:  #2439

### DIFF
--- a/packages/components/dialog/src/DialogContent.styles.tsx
+++ b/packages/components/dialog/src/DialogContent.styles.tsx
@@ -1,7 +1,12 @@
 import { cva, VariantProps } from 'class-variance-authority'
 
 export const dialogContentStyles = cva(
-  ['z-modal flex flex-col bg-surface group', 'focus-visible:outline-none focus-visible:u-ring'],
+  [
+    'z-modal flex flex-col bg-surface group',
+    'focus-visible:outline-none focus-visible:u-ring',
+    '[&:not(:has(footer))]:pb-lg',
+    '[&:not(:has(header))]:pt-lg',
+  ],
   {
     variants: {
       size: {


packages/components/dialog/src/DialogContent.styles.tsx
```diff
@@ -1,7 +1,12 @@
 import { cva, VariantProps } from 'class-variance-authority'
 
 export const dialogContentStyles = cva(
-  ['z-modal flex flex-col bg-surface group', 'focus-visible:outline-none focus-visible:u-ring'],
+  [
+    'z-modal flex flex-col bg-surface group',
+    'focus-visible:outline-none focus-visible:u-ring',
+    '[&:not(:has(footer))]:pb-lg',
+    '[&:not(:has(header))]:pt-lg',
```
soykje, line 8: If I understand correctly, this would be broken if we use another tag element than `header|footer`? Why not simply using `first-child|last-child` instead? Might be more robust here, wdyt?

acd02: Since `Dialog.Header` and `Dialog.Footer`serve as semantic header and footer elements respectively, it's unlikely that we would use div or other HTML elements instead

soykje: Even with `asChild` API?

Powerplex (approved):

packages/components/dialog/src/DialogContent.styles.tsx
Powerplex: If we need to detect that header/footer are not used in the future and the css selector is not enough, we could use the context

PR #2456, opened 2024-10-07
fix(dialog): do not close dialog when foreground elements are clicked

Powerplex:
**TASK**: [SPA-193](https://jira.ets.mpi-internal.com/browse/SPA-193)

Dialog/Drawer have a focus trap by default. It applies `pointer-events-none` on the whole page in the background to disabled clicks.

But some components might be displayed on top of dialogs and have a stronger `z-index` (ex: `Snackbar`, `Popover`). Such components must remain interactive, so they have `pointer-events-auto` on them.

Let's say I am inside a dialog:

1. I click somewhere inside the dialog that triggers a snackbar.
2. I click the close button of the snackbar displayed.
3. It closes the snackbar, **but ALSO the dialog.** (we don't want that).

Here are the z-index system of Spark: 
```
zIndex: {
    hide: '-1',
    base: '0',
    raised: '1',
    dropdown: '1000',
    sticky: '1100',
    overlay: '1300',
    modal: '1400', // dialog/drawer are using this one
    popover: '1500', 
    skipLink: '1600',
    toast: '1700',
    tooltip: '1800',
  },
```

As you can see `popover` and `toast` (snackbar) elements are on top. Such elements should not close the dialog when interacted with.

### Types of changes
- [x] 🧠 Refactor
- [x] 💄 Styles

### DIFF
--- a/packages/components/dialog/src/DialogContent.tsx
+++ b/packages/components/dialog/src/DialogContent.tsx
@@ -13,7 +13,14 @@ export interface ContentProps extends RadixDialog.DialogContentProps, DialogCont
 
 export const Content = forwardRef(
   (
-    { children, className, isNarrow = false, size = 'md', ...rest }: ContentProps,
+    {
+      children,
+      className,
+      isNarrow = false,
+      size = 'md',
+      onInteractOutside,
+      ...rest
+    }: ContentProps,
     ref: Ref<HTMLDivElement>
   ): ReactElement => {
     const { setIsFullScreen } = useDialog()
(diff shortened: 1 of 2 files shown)


soykje:
As discussed this may allow dark pattern (in terms of a11y). I'm not sure we should have this in our code.

soykje (approved):
As discussed this change is risky IMHO. Thus, as it might be still necessary to have this, we will have to prioritize documentation on both dev and design sides, to help them avoid dark patterns :+1:

PR #2488, opened 2024-10-22
feat(skeleton): add skeleton items with animation

soykje:
## TYPE(SCOPE): TITLE

**TASK**:


packages/components/skeleton/src/SkeletonItem.styles.ts
```diff
@@ -0,0 +1,37 @@
+import { cva, VariantProps } from 'class-variance-authority'
+
+export const skeletonItemStyles = cva(['bg-neutral/dim-4', 'min-h-lg min-w-lg'], {
```
Powerplex, line 3: `min-h-lg min-w-lg` -> `min-size-lg`

packages/components/skeleton/src/SkeletonItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+export const SkeletonRectangle = ({ width = '100%', height, ...rest }: SkeletonRectangleProps) => {
+  return (
+    <SkeletonItem
+      style={
+        {
+          '--skeleton-rect-width': getSizeValue(width),
+          '--skeleton-rect-height': getSizeValue(height),
+        } as CSSProperties
+      }
+      className="h-[--skeleton-rect-height] w-[--skeleton-rect-width]"
```
Powerplex, line 57: Maybe use inline styles for this instead of TW classNames ? (like we advise our consumers to do when they use arbitrary value)

packages/components/skeleton/src/SkeletonItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+)
+
+export const SkeletonLine = ({
+  lines = 1,
+  gap: gapProp,
+  alignment = 'start',
+  className,
+  ...rest
+}: SkeletonLineProps) => {
+  const gap = gapProp || (lines > 1 ? 'md' : undefined)
```
Powerplex, line 82: `const gap = gapProp || 'md'`

You don't need to check for the number of lines, because a flex-gap applied when a single element is rendered won't be visible anyway (it avoids a runtime check)

packages/components/skeleton/src/SkeletonGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Sets the main axis direction.
+   * @default row
+   */
+  direction?: 'row' | 'column'
+  /**
+   * Sets the default alignement.
+   * @default start
+   */
+  alignment?: 'start' | 'center' | 'end'
```
Powerplex, line 17: In other components, we named this prop `align`

packages/components/skeleton/src/Skeleton.styles.ts
```diff
@@ -0,0 +1,19 @@
+import { cva, VariantProps } from 'class-variance-authority'
+
+export const skeletonStyles = cva(['flex flex-row !flex-wrap'], {
+  variants: {
+    isAnimated: {
+      true: [
+        '[mask-image:linear-gradient(90deg,#000_40%,rgba(0,0,0,.3)_60%,#000_100%)]',
+        '[mask-size:200%_100%]',
+        'animate-shimmer',
```
Powerplex, line 9: You can add `motion-reduce:!animate-none` for a11y :)

packages/components/skeleton/src/SkeletonItem.tsx
soykje: This is what I did initially. But then, when I wanted to overwrite the value there were priority issues :confused:

packages/components/skeleton/src/Skeleton.styles.ts
soykje: Don't we add this globally? Wasn't aware of that, thx!

packages/components/skeleton/src/SkeletonItem.styles.ts
soykje: Are you sure this class exists? I can't make it work :thinking:

Powerplex (approved):

PR #2489, opened 2024-10-23
docs: updated carddesign of Getting Started section

Powerplex:
**TASK**: [SPA-352](https://jira.ets.mpi-internal.com/browse/SPA-352)

Design update for readability

### Types of changes
- [x] 🧾 Documentation

### Screenshots - Animations
Capture d’écran 2024-10-23 à 16 13 22


soykje:
@Powerplex As discussed with @acd02 I'm not convinced we should add this content in our Storybook. I'd rather keep it in the Zeroheight and maybe work on reinforcing the links between both, wdyt?

PR #2554, opened 2025-01-16
feat(visually-hidden): added asChild prop to VisuallyHidden

Powerplex:
**TASK**: [SPA-485](https://jira.ets.mpi-internal.com/browse/SPA-485)

Missing `asChild` prop on `VisuallyHidden` to make it polymorphic. Until now it was a `span` and could not be any other html tag.

### Types of changes

### DIFF
--- a/packages/components/visually-hidden/package.json
+++ b/packages/components/visually-hidden/package.json
@@ -23,6 +23,9 @@
   "scripts": {
     "build": "vite build"
   },
+  "dependencies": {
+    "@spark-ui/slot": "^8.0.2"
+  },
   "peerDependencies": {
     "react": "^19.0",
     "react-dom": "^19.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -30460,6 +30460,9 @@
       "name": "@spark-ui/visually-hidden",
       "version": "8.0.3",
       "license": "MIT",
+      "dependencies": {
+        "@spark-ui/slot": "^8.0.2"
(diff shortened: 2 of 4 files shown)


oudoulj (approved):

packages/components/visually-hidden/package.json
```diff
@@ -23,6 +23,9 @@
   "scripts": {
     "build": "vite build"
   },
+  "dependencies": {
+    "@spark-ui/slot": "^8.0.2"
```
acd02, line 27:
```suggestion
    "@spark-ui/slot": "8.0.2"
```

Powerplex: Are you sure ? This is like this in every package.


https://github.com/user-attachments/assets/68b9955b-d60d-4dbf-b738-5ab9a22ed90e

acd02: That should not be the case imho, as it may introduce different users with varying versions of those dependencies installed on their machines, making it difficult to debug in the event of bugs.

See 👉https://github.com/adevinta/spark/wiki/Agreements#managing-dependencies-within-spark-components

Powerplex: Are you ok if I replace everywhere then but in a separate PR ? I have a feeling that we used to remove the `^` in the passed and that it is possibly the fixed versioning that automatically added it back everywhere ?

PR #2586, opened 2025-02-17
fix(chip): clear button cursor pointer

andresin87:
## fix(chip): clear button cursor pointer

**TASK**: 

### Types of changes


Powerplex (approved):

Powerplex (changes requested):
It means both `cursor-pointer` and `cursor-not-allowed` can be applied at the same time. This makes the behavior unpredictable because we don't have control over the order in which classes are defined in tailwind css output.

PR #2716, opened 2025-05-16
feat(components): improves stepper empty value behavior with min/max limits

soykje:
When `Stepper` inits with empty value and min|max values are set to limit accepted range to strictly positive or negative values, we shouldn't allow increment|decrement actions.

For example, if we have a controlled input with empty value that has a `minValue` set to `0`, we shouldn't have the possibility to decrement as it would lead to out of range value (as reference starting value is always 0). In the same way we want to adapt this logic when `maxValue` is set to `0` : then increment button should be disabled.

### Types of changes
- [x] 🧾 Documentation
- [x] 🧪 Test
- [x] 🧠 Refactor

### DIFF
--- a/packages/components/src/stepper/Stepper.doc.mdx
+++ b/packages/components/src/stepper/Stepper.doc.mdx
@@ -32,7 +32,7 @@ export default () => (
     <Stepper.Input />
     <Stepper.IncrementButton />
   </Stepper>
-);
+)
 ```
 
 ## Examples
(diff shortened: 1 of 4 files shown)


packages/components/src/stepper/Stepper.doc.mdx
```diff
@@ -79,6 +79,8 @@ Use `minValue|maxValue` props to limit the entered value to a specific range. Ra
 
 Increment and decrement buttons will be disabled when the value is within one step value from the bounds.
 
+If the value is empty (unset or cleared) and `minValue|maxValue` set to limitate value to strictly positive or negative values then increment and decrement buttons will be disabled accordingly.
```
oudoulj, line 82:
```suggestion
If the value is empty (unset or cleared) and `minValue|maxValue` set to limit value to strictly positive or negative values then increment and decrement buttons will be disabled accordingly.
```

oudoulj (approved):

PR #2743, opened 2025-06-06
fix(components): update design checkbox

Powerplex:
Closes JIRA [LBCSPARK-259](https://jira.ets.mpi-internal.com/browse/LBCSPARK-259)
Closes JIRA [LBCSPARK-260](https://jira.ets.mpi-internal.com/browse/LBCSPARK-260)

On both `Radio` and `Checkbox`:
- only `basic` and `error` intents remains documented (NOTE: the other intents are not removed to avoid a breaking change, but should no longer be used)
- `error` intent now display a red outline even in unchecked state (vs "outline" color before the changes)
- the `reverse` prop documentation has been removed, as for `a11y` this is not something we want to encourage.

### Types of changes
- [x] 🧾 Documentation
- [x] 💄 Styles

### Screenshots - Animations

CHECKBOX
RADIO

### DIFF
--- a/packages/components/src/checkbox/Checkbox.doc.mdx
+++ b/packages/components/src/checkbox/Checkbox.doc.mdx
@@ -80,20 +80,6 @@ Use the `disabled` prop to indicate that the checkbox is disabled.
 
 <Canvas of={stories.Disabled} />
 
-### Reverse
-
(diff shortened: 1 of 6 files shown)


Bibisirius:
✅ Review design : Validated

Bibisirius:
✅ Review design : Validated

PR #3003, opened 2026-04-09
docs(components): document disabled/readOnly behavior with FormField in FileUpload

musenga17:
document disabled/readOnly behavior with FormField in FileUpload

Add documentation clarifying that when `FileUpload` is used inside a `FormField`, the `disabled `and `readOnly` props must be set on the `FormField` component rather than on `FileUpload` itself, as `FormField` automatically propagates these states.                                                    
- Add warning callouts in "Disabled state" and "Read-only state" sections                                                         
- Add info callout in "Form field" section explaining state management

### DIFF
--- a/packages/components/src/file-upload/FileUpload.doc.mdx
+++ b/packages/components/src/file-upload/FileUpload.doc.mdx
@@ -262,11 +262,23 @@ Use `FileUpload.Context` to access `acceptedFiles` and `rejectedFiles`, then use
 
 Use the `disabled` prop to prevent user interaction with the file upload component. When disabled, users cannot add, remove, or modify files. The component is visually disabled with reduced opacity and a "not-allowed" cursor. Existing files are still displayed but cannot be modified. Disabled fields do not participate in form submission.
 
+<Callout kind="warning" marginY="large">
+  <p>
+    <strong>Important:</strong> When using `FileUpload` inside a `FormField`, you must set the `disabled` prop on the `FormField` component rather than on `FileUpload` itself. The `FormField`'s `disabled` state will be automatically propagated to the `FileUpload`.
+  </p>
+</Callout>
+
 <Canvas of={stories.Disabled} />
 
 ### Read-only state
 
 Use the `readOnly` prop to set the file upload to read-only mode. Similar to disabled, users cannot add, remove, or modify files, but the component maintains a more subtle visual appearance (no opacity reduction). Existing files are displayed but cannot be modified. Read-only fields participate in form submission.
+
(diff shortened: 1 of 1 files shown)


Powerplex:
Thanks.

The issue is that this is valid for every FormField compatible components, so I preferred to document it in `FormField` directly. This is also valid for aria attributes and state (error/success/warning) that is passed to the input in many cases.

I prefer it remains implicit.

The full list of compatible components is: 
```

packages/components/src/checkbox (and CheckboxGroup)
packages/components/src/combobox
packages/components/src/file-upload
packages/components/src/input (and InputGroup)
packages/components/src/input
packages/components/src/input-otp
packages/components/src/textarea
packages/components/src/radio-group
packages/components/src/rating/Rating.tsx:
packages/components/src/segmented-control
packages/components/src/select
packages/components/src/slider
packages/components/src/stepper
packages/components/src/switch

```

If we add it to `FileUpload` then we must add it for every other component too. But I think I prefer if the doc regarding this is centralized in the FormField doc page instead (we could add Disabled/ReadOnly examples on this page ? https://sparkui.vercel.app/?path=/docs/components-formfield--docs#examples) What do you think ?

musenga17:
> Thanks.
> 
> The issue is that this is valid for every FormField compatible components, so I preferred to document it in `FormField` directly. This is also valid for aria attributes and state (error/success/warning) that is passed to the input in many cases.
> 
> I prefer it remains implicit.
> 
> The full list of compatible components is:
> 
> ```
> 
> packages/components/src/checkbox (and CheckboxGroup)
> packages/components/src/combobox
> packages/components/src/file-upload
> packages/components/src/input (and InputGroup)
> packages/components/src/input
> packages/components/src/input-otp
> packages/components/src/textarea
> packages/components/src/radio-group
> packages/components/src/rating/Rating.tsx:
> packages/components/src/segmented-control
> packages/components/src/select
> packages/components/src/slider
> packages/components/src/stepper
> packages/components/src/switch
> ```
> 
> If we add it to `FileUpload` then we must add it for every other component too. But I think I prefer if the doc regarding this is centralized in the FormField doc page instead (we could add Disabled/ReadOnly examples on this page ? https://sparkui.vercel.app/?path=/docs/components-formfield--docs#examples) What do you think ?

Thanks for the clarification, that makes total sense.

I agree that keeping this behavior implicit and documenting it centrally in the FormField page is a better approach for consistency across all compatible components.

Adding Disabled/ReadOnly examples there sounds like a great idea 👍

I’m good with this direction!

